PR #5, opened 2020-06-29
Add Dir::is_dir and refactor Dir::create_dir_all

kubkon:
This commit adds `Dir::is_dir` and refactors `Dir::create_dir_all`
which makes use of it internally. This commit also enables several
tests testing `create_dir_all` along both positive and negative
execution paths.

### DIFF
--- a/tests/fs.rs
+++ b/tests/fs.rs
@@ -31,47 +31,6 @@ use cap_std::os::windows::fs::{symlink_dir, symlink_file};
 use cap_std::sys::fs::symlink_junction;
 */
 
-macro_rules! check {
-    ($e:expr) => {
-        match $e {
-            Ok(t) => t,
-            Err(e) => panic!("{} failed with: {}", stringify!($e), e),
-        }
-    };
-}
-
-#[cfg(windows)]
-macro_rules! error {
-    ($e:expr, $s:expr) => {
-        match $e {
-            Ok(_) => panic!("Unexpected success. Should've been: {:?}", $s),
-            Err(ref err) => assert!(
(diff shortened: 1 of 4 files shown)


tests/fs.rs
```diff
@@ -535,18 +535,18 @@ fn mkdir_path_already_exists_error() {
     assert_eq!(e.kind(), ErrorKind::AlreadyExists);
 }
 
-/*
 #[test]
 fn recursive_mkdir() {
     let tmpdir = tmpdir();
     let dir = "d1/d2";
     check!(tmpdir.create_dir_all(dir));
-    assert!(dir.is_dir())
```
sunfishcode, line 544: This should translate into `assert!(tmpdir.is_dir(dir));`

tests/fs.rs
```diff
@@ -535,18 +535,18 @@ fn mkdir_path_already_exists_error() {
...
+    assert!(tmpdir.is_dir("d1"));
+    let dir = check!(tmpdir.open_dir("d1"));
+    assert!(dir.is_dir("d2"));
+    assert!(tmpdir.is_dir("d1/d2"));
```
sunfishcode, line 546: These extra checks are good, though would it be inconvenient to add them in a separate file? That way we can keep `fs.rs` as close to a direct translation of Rust's `fs.rs` as possible.

kubkon:
OK, @sunfishcode, I've now addressed your comments in e5a077a. Have a look and lemme know what you think. I've had to do some small refactoring to get `check!` and other macros available across the entire `tests` crate.

sunfishcode:
Looks great!

PR #6, opened 2020-07-01
Implement OpenOptionsExt for yanix platforms.

sunfishcode:
And stub out Windows support. Also, implement `open_file` and
`create_file` in terms of `open_file_with`, for less unsafe code and
fewer magic constants.

### DIFF
--- a/src/sys/yanix/fs/dir.rs
+++ b/src/sys/yanix/fs/dir.rs
@@ -3,6 +3,7 @@ use crate::{
     os::unix::net::{UnixDatagram, UnixListener, UnixStream},
 };
 use std::{
+    convert::TryInto,
     fs, io,
     os::unix::{
         fs::OpenOptionsExt,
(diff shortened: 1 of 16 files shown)


src/sys/yanix/fs/dir.rs
```diff
@@ -69,27 +60,31 @@ impl Dir {
...
+            & !OFlag::ACCMODE;
         unsafe {
             let fd = openat(
                 self.std_file.as_raw_fd(),
                 path,
                 oflags,
-                Mode::from_bits(0o666).unwrap(),
+                #[allow(clippy::useless_conversion)]
+                Mode::from_bits(options.ext.mode.try_into().expect("unrecognized Mode bits"))
+                    .expect("unrecognized Mode bits"),
```
kubkon, line 72: Should we panic here? Or is a situation where the user specifies wrong mode bits possible, and which they'd like to recover from?

sunfishcode: Looks like libstd silently ignores unrecognized mode bits. I'll update to match.

sunfishcode: Done!

PR #8, opened 2020-07-01
Add Dir::canonicalize and read_link (in *nix)

kubkon:
Canonicalization is handled by `PathNormalizer` struct which ensures we
only ever return a relative path to the root Dir handle without
moving beyond it. It is possible since `PathNormalizer` is an adaption
of `path::get` from wasi-common, and essentially emulates Capsicum.

Builds on #7


kubkon:
As discussed offline, closing as it’s superseded by #9. Instead will backport the approach to the testing crate after #9 lands.

PR #9, opened 2020-07-01
Implement sandboxing for `open`.

sunfishcode:
Reorganize the codebase to better share code between cap-async-std and
cap-std, by factoring out a cap-primitives crates that they can both
depend on. In cap-primitives, implement sandboxing for `open`.

The crates are now part of a workspace, so they can all be built
together.

This PR is based on #7.


cap-primitives/src/std/fs/get_path.rs
```diff
@@ -0,0 +1,35 @@
...
+    // alternatives. If a better method is invented, it should be used
+    // instead.
+    let mut buf = vec![0; libc::PATH_MAX as usize];
+    let n = unsafe { libc::fcntl(file.as_raw_fd(), libc::F_GETPATH, buf.as_ptr()) };
+    if n == -1 {
+        return None;
+    }
+    let l = buf.iter().position(|&c| c == 0).unwrap();
+    buf.truncate(l as usize);
+    buf.shrink_to_fit();
```
kubkon, line 28: This is super unfortunate and alas I've not found a better way at solving this on Darwin... In fact, I've got pretty much the exact same solution over in Zig: [zig#5701](https://github.com/ziglang/zig/pull/5701/files#diff-5a679ae5685f723b7d3fec367ffbeab6R3951).

cap-primitives/src/yanix/linux/fs/open.rs
```diff
@@ -0,0 +1,142 @@
...
+        // Test if `openat2` is supported. If so, we can use `open_openat2`.
+        // Otherwise, fall back to `open_manually`.
+        unsafe {
+            if let -1 = libc::syscall(
+                SYS_OPENAT2,
+                -1,
+                ptr::null::<libc::c_void>(),
+                ptr::null::<libc::c_void>(),
+                0,
+            ) {
```
kubkon, line 109: This looks like a perfect spot to reuse Alex's idea of having a global atomic bool set to `true` if the syscall doesn't exist (and falling back to an alternative), or `false` otherwise. This should be thread safe, plus make it a little bit more maintainable in the long run. This would then look something like:

```rust
static INVALID: AtomicBool = AtomicBool::new(false);
if !INVALID.load(SeqCst) {
    let rc = unsafe {
        libc::syscall(
            libc::SYS_openat2,
            start.as_raw_fd(),
            path_cstr.as_ptr(),
            &open_how,
            SIZEOF_OPEN_HOW,
       )
    };
    match rc {
        -1 => match io::Error::last_os_error().raw_os_error().unwrap() {
            libc::ENOSYS => INVALID.store(true, SeqCst);
        }
        // ...
    }
}

open_manually_wrapper(start, path, options)
```

FWIW though, I'm happy to follow up with this one in a subsequent PR.

cap-primitives/src/std/fs/open.rs
```diff
@@ -0,0 +1,85 @@
...
+/// Perform an `openat`-like operation, ensuring that the resolution of the path
+/// never escapes the directory tree rooted at `start`.
+pub fn open(start: &fs::File, path: &Path, options: &OpenOptions) -> io::Result<fs::File> {
+    // Call `open`.
+    let result = open_impl(start, path, options);
+
+    // Do an unsandboxed lookup and check that we found the same result.
+    // TODO: This is a racy check, though it is useful for testing and fuzzing.
+    #[cfg(debug_assertions)]
+    match open_unchecked(
```
kubkon, line 19: Will this even work in WASI?

cap-primitives/src/std/fs/open_options.rs
```diff
@@ -128,7 +135,7 @@ impl std::os::unix::fs::OpenOptionsExt for OpenOptions {
 
 #[cfg(target_os = "vxworks")]
 impl std::os::vxworks::fs::OpenOptionsExt for OpenOptions {
-    fn mode(&mut self, mode: libc::mode_t) -> &mut Self {
+    fn mode(&mut self, mode: u32) -> &mut Self {
```
kubkon, line 138: Oh, how come?

cap-primitives/src/std/fs/open.rs
sunfishcode: My plan for WASI is to have it skip this code altogether; details in an upcoming patch :-).

cap-primitives/src/std/fs/open_options.rs
sunfishcode: That's how the trait is defined :shrug:

kubkon: Fair enough!

cap-primitives/src/std/fs/maybe_owned_file.rs
```diff
@@ -0,0 +1,22 @@
+use std::{fs, io};
+
+pub(crate) enum MaybeOwnedFile<'borrow> {
+    Owned(fs::File),
+    Borrowed(&'borrow fs::File),
+}
```
kubkon, line 6: Hmm, I'm wondering, could we somehow re-use the concept of clone-on-write smart pointer here? Just an idea. I'm not that familiar with `Cow` myself TBH, but it looks familiar.

cap-primitives/src/yanix/linux/fs/open.rs
sunfishcode: Cool, I think that'd be a good fit here. If you could add this in a follow up that'd be great!

cap-primitives/src/std/fs/maybe_owned_file.rs
sunfishcode: It is very similar; the difference is that `Cow` requires its `T` to implement `ToOwned`, which requires an infallible way to go from `&T` to `T`. `std::fs::File` can't do that because cloning a `File` is fallible.

kubkon: Ah, gotcha!

kubkon: If we ever find ourselves in the need of this fallible idiom, I guess we could extract it into a generic `TryCow<T: Fallible>` or some such. But until then, this looks good!

sunfishcode: I think we want something like `Cow` but without the whole "copy on write" thing going on ;-).

PR #10, opened 2020-07-02
Begin porting to wasm32-wasi.

sunfishcode:
This needs updates in yanix and libc before it'll be full buildable,
but this gets things started.


kubkon:
Could you rebase so that we could merge this in?

sunfishcode:
Rebased!

PR #13, opened 2020-07-05
Refactor openat2

kubkon:
This PR swaps the use of `lazy_static` to store the pointer to
a function (which can either be `openat2` or `open_manully_wrapper`),
re-use Alex's idea to store a predicate in `AtomicBool` which is set
to `false` initially. Upon calling `openat2` for the first time, if
the host returns `ENOSYS`, swap the bool for `false` and from that
point on, always fall back to `open_manually_wrapper`.

It also fixes debug assert in `openat2_or_open_manually` function, and
adds host-dependent `Dir::open_file` since we need to be able
to tweak mode bits on Linux when using `openat2` syscall. Otherwise,
we'll run into `EINVAL` errnos, since `openat2` is very strict about
passing incorrect `open_how` structure.

EDIT: ~~Unfortunately, `cap_smoke_test` doesn't yet pass on a kernel with `openat2`
provided, but I'll try to debug some more, and see if I can fix it. Hence,
for the time being, I'm marking the PR as WIP.~~ All tests are green on Linux with kernel `>=5.6`.
As discussed offline, I'm currently in the process of figuring out if we could source a CI
which supports Linux images with kernel `5.6` or above.

### DIFF
--- a/cap-primitives/src/yanix/linux/fs/open.rs
+++ b/cap-primitives/src/yanix/linux/fs/open.rs
@@ -50,7 +50,7 @@ pub(crate) fn open_impl(
         // Until then, since `O_TMPFILE := 0x20000000 | libc::O_DIRECTORY`,
         // we need to compare for bit equality.
         let mode = if (oflags.bits() & libc::O_CREAT == libc::O_CREAT)
-            || (oflags.bits() & libc::O_TMPFILE == libc::O_TMPFILE) != 0
+            || (oflags.bits() & libc::O_TMPFILE == libc::O_TMPFILE)
         {
             options.ext.mode
         } else {
--- a/cap-primitives/src/yanix/linux/fs/open.rs
+++ b/cap-primitives/src/yanix/linux/fs/open.rs
@@ -45,9 +45,13 @@ pub(crate) fn open_impl(
     static INVALID: AtomicBool = AtomicBool::new(false);
     if !INVALID.load(Relaxed) {
         let oflags = compute_oflags(options);
-        
+
         // TODO use `yanix::file::OFlags` when `TMPFILE` is introduced
-        let mode = if oflags.bits() & (libc::O_CREAT | libc::O_TMPFILE) != 0 {
+        // Until then, since `O_TMPFILE := 0x20000000 | libc::O_DIRECTORY`,
+        // we need to compare for bit equality.
+        let mode = if (oflags.bits() & libc::O_CREAT == libc::O_CREAT)
+            || (oflags.bits() & libc::O_TMPFILE == libc::O_TMPFILE) != 0
+        {
             options.ext.mode
         } else {
             0
--- a/cap-primitives/src/yanix/linux/fs/open.rs
+++ b/cap-primitives/src/yanix/linux/fs/open.rs
@@ -78,7 +78,7 @@ fn openat2_or_open_manually(
                         errno => return other_error(errno),
                     },
                     ret => {
-                        let fd = ret as RawFd;
+                        let file = fs::File::from_raw_fd(ret as RawFd);
 
                         #[cfg(debug_assertions)]
                         {
@@ -93,12 +93,12 @@ fn openat2_or_open_manually(
                             )
                             .expect("open_manually failed when open_openat2 succeeded");
                             debug_assert!(
-                                is_same_file(start, &check)?,
+                                is_same_file(&file, &check)?,
                                 "open_manually should open the same inode as open_openat2"
                             );
                         }
 
-                        return Ok(fs::File::from_raw_fd(fd));
+                        return Ok(file);
                     }
                 }
             }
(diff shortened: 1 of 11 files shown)


cap-primitives/src/yanix/linux/fs/open.rs
```diff
@@ -35,57 +35,81 @@ struct OpenHow {
...
+                    path_cstr.as_ptr(),
+                    &open_how,
+                    SIZEOF_OPEN_HOW,
+                ) {
+                    -1 => match io::Error::last_os_error().raw_os_error().unwrap() {
+                        libc::EAGAIN => continue,
+                        libc::EXDEV => {
+                            if path.is_absolute() {
+                                return absolute_path();
+                            }
```
sunfishcode, line 79: This unfortunately doesn't handle the case where `path` is a symlink to an absolute path, so it'll still print a different message than `open_manually` in that case. I like the overall direction of this patch, moving toward making the error messages consistent between the different implementations, so it seems like we should take it even further and just have one message for both ".." and "/" escape attempts.

src/fs/dir.rs
```diff
@@ -48,7 +48,9 @@ impl Dir {
     /// [`std::fs::File::open`]: https://doc.rust-lang.org/std/fs/struct.File.html#method.open
     #[inline]
     pub fn open_file<P: AsRef<Path>>(&self, path: P) -> io::Result<File> {
-        self.open_file_with(path, OpenOptions::new().read(true))
+        // We're delegating to a sys-dependent open_file due to `openat2` being really strict
+        // about mixing oflags and mode bits.
+        self.sys.open_file(path.as_ref())
```
sunfishcode, line 53: Instead of bubbling this up, could we handle this locally within the `openat2`-specific code? That is, instead of it doing:
```rust
    let mode = options.ext.mode;
```
have it check if the flags contain O_CREAT or O_TMPFILE and set the mode to 0 if not?

cap-primitives/src/yanix/linux/fs/open.rs
```diff
@@ -35,57 +35,81 @@ struct OpenHow {
...
-    let mode = options.ext.mode;
+/// Call the `openat2` system call. If the syscall is unavailable, mark it so for future
+/// calls, and fallback to `open_manually_wrapper`
+fn openat2_or_open_manually(
+    start: &fs::File,
+    path: &Path,
+    options: &OpenOptions,
+) -> io::Result<fs::File> {
+    static INVALID: AtomicBool = AtomicBool::new(false);
+    if !INVALID.load(SeqCst) {
```
sunfishcode, line 46: Can we use `Relaxed` instead of `SeqCst` here and in the `store`? We don't care about ordering here, and this is consistent with what [libstd does](https://github.com/rust-lang/rust/blob/master/src/libstd/sys/unix/rand.rs#L45) in various places.

cap-primitives/src/yanix/linux/fs/open.rs
```diff
@@ -95,37 +119,13 @@ fn open_openat2(start: &fs::File, path: &Path, options: &OpenOptions) -> io::Res
...
-}
-
 #[inline]
 pub(crate) fn open_impl(
     start: &fs::File,
     path: &Path,
     options: &OpenOptions,
 ) -> io::Result<fs::File> {
-    OPEN(start, path, options)
+    openat2_or_open_manually(start, path, options)
```
sunfishcode, line 128: `open_impl` here is now just a wrapper around `openat2_or_open_manually` that doesn't do anything on its own. Would it make sense to make `openat2_or_open_manually` be `pub(crate)`, and then have the mod.rs file do `use openat2_or_open_manually as open_impl`?

cap-primitives/src/yanix/linux/fs/open.rs
kubkon: Makes sense, thanks!

cap-primitives/src/yanix/linux/fs/open.rs
kubkon: Hmm, I like the direction of this, but I think it would be even better if we simply renamed `openat2_or_open_manually` to `open_impl`. This way, we don't have to tweak any other bit which uses `fs::open_impl`. Otherwise, since `openat2_or_open_manually` already resides in `fs` module, we couldn't simply re-export it as `open_impl` in `fs` mod in `mod.rs`, since `fs` is already defined. What do you think?

sunfishcode: Sounds good to me!

kubkon:
OK, once the tests pass, I think we're ready to go with this one.

src/fs/dir.rs
kubkon: Done, albeit in a very roundabout fashion since `yanix` crate doesn't expose `O_TMPFILE` yet. This will need to get fixed on the `yanix` side. I left a TODO as a reminder what needs to be done.

sunfishcode:
Looks great!

PR #15, opened 2020-07-06
Hide use of libc for error handling in open_unchecked

kubkon:
This way we stay host-independent which is quite important for
Windows compat.

### DIFF
--- a/cap-primitives/src/std/fs/open.rs
+++ b/cap-primitives/src/std/fs/open.rs
@@ -53,6 +53,7 @@ pub fn open(start: &fs::File, path: &Path, options: &OpenOptions) -> io::Result<
             Err(result_error) => match result_error.kind() {
                 io::ErrorKind::PermissionDenied | io::ErrorKind::InvalidInput => (),
                 _ => {
+                    let unchecked_error = io::Error::from(unchecked_error);
                     assert_eq!(result_error.to_string(), unchecked_error.to_string());
                     assert_eq!(result_error.kind(), unchecked_error.kind());
                 }
--- a/cap-primitives/src/std/fs/open.rs
+++ b/cap-primitives/src/std/fs/open.rs
@@ -54,10 +54,8 @@ pub fn open(start: &fs::File, path: &Path, options: &OpenOptions) -> io::Result<
                 io::ErrorKind::PermissionDenied | io::ErrorKind::InvalidInput => (),
                 _ => {
                     let unchecked_error = match unchecked_error {
-                        OpenUncheckedError::Io(err) => err,
-                        OpenUncheckedError::Symlink => {
-                            io::Error::new(io::ErrorKind::Other, "unexpected ELOOP or EMLINK")
-                        }
+                        OpenUncheckedError::Other(err) => err,
+                        OpenUncheckedError::Symlink(err) => err,
                     };
                     assert_eq!(result_error.to_string(), unchecked_error.to_string());
                     assert_eq!(result_error.kind(), unchecked_error.kind());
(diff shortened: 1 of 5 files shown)


cap-primitives/src/std/fs/open_manually.rs
```diff
@@ -201,3 +197,26 @@ fn escape_attempt() -> io::Result<fs::File> {
...
+
+#[cold]
+fn symlink_disallowed() -> io::Result<fs::File> {
+    Err(OpenUncheckedError::SymlinkDisallowed.into())
+}
+
+#[derive(Debug)]
+pub(crate) enum OpenUncheckedError {
+    Io(io::Error),
+    SymlinkDisallowed,
```
sunfishcode, line 209: Looking at this more closely, the code returns `SymlinkDisallowed` isn't the code that checks whether they're allowed or not. That check happens at the next level up. So I think we can rename this to just `Symlink` or `SymlinkError`.

cap-primitives/src/yanix/common/fs/open_unchecked.rs
```diff
@@ -27,8 +24,15 @@ pub(crate) fn open_unchecked(
...
+    let err = unsafe {
+        match openat(start.as_raw_fd(), path, oflags, mode) {
+            Ok(fd) => return Ok(fs::File::from_raw_fd(fd)),
+            Err(err) => err,
+        }
+    };
+    match err.raw_os_error() {
+        Some(libc::ELOOP) | Some(libc::EMLINK) => Err(OpenUncheckedError::SymlinkDisallowed),
+        None => Err(OpenUncheckedError::Io(err)),
+        Some(e) => Err(OpenUncheckedError::Io(io::Error::from_raw_os_error(e))),
```
sunfishcode, line 36: These two lines can just be `_ => OpenUncheckedError::Io(err),`, so that we pass through any extra error context that might be present in `err`.

cap-primitives/src/std/fs/open_manually.rs
```diff
@@ -141,18 +141,14 @@ pub(crate) fn open_manually(
...
-                        Some(libc::ELOOP) | Some(libc::EMLINK) => {
-                            let destination =
-                                resolve_symlink_at(base.as_file(), &one, symlink_count)?;
-                            components
-                                .extend(destination.components().map(to_owned_component).rev());
-                        }
-                        _ => return Err(e),
-                    },
+                    Err(OpenUncheckedError::Symlink) if use_options.nofollow => {
+                        return symlink_disallowed()
```
sunfishcode, line 145: Having the original `io::Error` would be useful here too, so that we can forward on the OS's original error message.

cap-primitives/src/std/fs/open.rs
```diff
@@ -53,6 +53,12 @@ pub fn open(start: &fs::File, path: &Path, options: &OpenOptions) -> io::Result<
             Err(result_error) => match result_error.kind() {
                 io::ErrorKind::PermissionDenied | io::ErrorKind::InvalidInput => (),
                 _ => {
+                    let unchecked_error = match unchecked_error {
+                        OpenUncheckedError::Io(err) => err,
+                        OpenUncheckedError::Symlink => {
+                            io::Error::new(io::ErrorKind::Other, "unexpected ELOOP or EMLINK")
```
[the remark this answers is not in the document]
kubkon: Agreed, this sounds like a good idea. I've done it now in 30e4ce5. We could in theory refactor `OpenUncheckedError` into a struct with some `kind` enum since currently both variants store `io::Error`, but I think it's OK as it is now since the error enum is so simple and not used in too many places.

sunfishcode:
Looks good!

PR #16, opened 2020-07-07
Implement `symlink_metadata`, `canonicalize`, and `symlink`.

sunfishcode:


cap-async-std/src/fs_utf8/dir_builder.rs
```diff
@@ -51,7 +51,7 @@ impl DirBuilder {
 #[cfg(unix)]
 impl async_std::os::unix::fs::DirBuilderExt for DirBuilder {
     #[inline]
-    fn mode(&mut self, mode: libc::mode_t) -> &mut Self {
+    fn mode(&mut self, mode: u32) -> &mut Self {
```
kubkon, line 54: Now that I started porting to Windows, this is actually a welcome change since we probably won't pull in `libc` when targeting Windows.

cap-primitives/src/std/fs/open_manually.rs
```diff
@@ -60,11 +60,28 @@ impl<'path_buf> CanonicalPath<'path_buf> {
...
+}
+
+impl<'path_buf> Drop for CanonicalPath<'path_buf> {
+    fn drop(&mut self) {
+        if let Some(path) = &mut self.path {
+            path.clear();
+            self.path = None;
         }
     }
 }
```
kubkon, line 87: I feel a bit uneasy about this. `self.path` is mutable reference so I'd imagine the owner should handle clearing, deallocating, etc. Besides, is this on-drop behaviour actually needed at all anywhere?

cap-primitives/src/yanix/common/fs/symlink_unchecked.rs
```diff
@@ -0,0 +1,24 @@
...
+use yanix::file::symlinkat;
+
+/// *Unsandboxed* function similar to `symlink`, but which does not perform sandboxing.
+pub(crate) fn symlink_unchecked(
+    old_path: &Path,
+    new_start: &fs::File,
+    new_path: &Path,
+) -> io::Result<()> {
+    // POSIX's `symlinkat` with an empty path returns `ENOENT`, so use "." instead.
+    let new_path = if new_path.components().next().is_none() {
```
kubkon, line 16: FWIW I think `new_path.as_os_str().is_empty()` communicates the intent better here.

src/fs_utf8/dir.rs
```diff
@@ -419,6 +455,52 @@ impl Dir {
...
+    ///
+    /// This corresponds to [`std::path::Path::exists`], but only
+    /// accesses paths relative to `self`.
+    ///
+    /// [`std::path::Path::exists`]: https://doc.rust-lang.org/std/path/struct.Path.html#method.exists
+    #[inline]
+    pub fn exists<P: AsRef<str>>(&self, path: P) -> bool {
+        let path = match from_utf8(&path) {
+            Ok(path) => path,
+            Err(_) => return false,
```
kubkon, line 469: I think it might useful explaining in the docs that if the `path` is not a valid UTF8, this will return `false`.

src/fs_utf8/dir.rs
```diff
@@ -419,6 +455,52 @@ impl Dir {
...
+    ///
+    /// This corresponds to [`std::path::Path::is_file`], but only
+    /// accesses paths relative to `self`.
+    ///
+    /// [`std::path::Path::is_file`]: https://doc.rust-lang.org/std/path/struct.Path.html#method.is_file
+    #[inline]
+    pub fn is_file<P: AsRef<str>>(&self, path: P) -> bool {
+        let path = match from_utf8(&path) {
+            Ok(path) => path,
+            Err(_) => return false,
```
kubkon, line 484: Same here.

src/fs_utf8/dir.rs
```diff
@@ -419,6 +455,52 @@ impl Dir {
...
+    /// This is similar to [`std::path::Path::is_dir`] in that it checks if `path` relative to `Dir`
+    /// is a directory. This function will traverse symbolic links to query information about the
+    /// destination file. In case of broken symbolic links, this will return `false`.
+    ///
+    /// [`std::path::Path::is_dir`]: https://doc.rust-lang.org/std/path/struct.Path.html#method.is_dir
+    #[inline]
+    pub fn is_dir<P: AsRef<str>>(&self, path: P) -> bool {
+        let path = match from_utf8(&path) {
+            Ok(path) => path,
+            Err(_) => return false,
```
kubkon, line 500: Same here.

cap-primitives/src/std/fs/open_manually.rs
sunfishcode: `canonicalize` can succeed even if the last component of the path is not accessible. `open_manually` will `Err` in that case, so `canonicalize` needs a separate way to determine if it got a complete path or not. Clearing the `canonical_path` is how the current patch indicates that the complete path is not available.

It comes from trying to keep `canonicalize`'s impact on `open_manually` minimal, because none of its other users care about the canonical path. With the current path, all they have to do is pass `None`. And then `canonicalize` passes a `&mut PathBuf` which then comes back with either the desired result or empty.

I've now added some comments to clarify this. Does this clarify what's going on?

src/fs_utf8/dir.rs
sunfishcode: There is a subtlety here: `path` is an `AsRef<str>`, so it's already guaranteed to be valid UTF-8. `fs_utf8` is using arf-strings to encode non-UTF-8 paths as valid UTF-8 strings. So it isn't "there might be a file there, but we're hiding it from you". You can always name and access that file, you just need to use the proper encoding (and that's what you're opting into by using the `fs_utf8` module :-)). `from_utf8` only fails if the arf-string encoding is invalid.

kubkon: That's fair enough. However, I'd still vote in favour of leaving a doc comment to remind the user about the possible cause of the failure. Having said that, perhaps this should be documented in the main `fs_utf8` crate/module's manifest instead of above every method? I know this should be obvious and this is what the user implicitly signs up for when using `fs_utf8`, but I've seen it countless of times when a tiny detail like this would go unnoticed just to rediscover this basic truth after hours of debugging.

cap-primitives/src/std/fs/open_manually.rs
kubkon: Yep, thanks!

In a future effort, perhaps it'd be worth refactoring it into a state machine or some such? This way, we could signal failure by encoding a state, and the user would inquire about the state to work out if the processing was successful or not? This way, we wouldn't have to touch the ref to `PathBuf` in `drop`.

src/fs_utf8/dir.rs
sunfishcode: Good idea. I've now added a paragraph to the module comment expalining what the `fs_utf8` module is all about.

kubkon: Nice! This is perfect, thanks!

PR #19, opened 2020-07-08
Enable FreeBSD coverage in srht

kubkon:
Also, tweak `MetadataExt` where `st_ino` on FreeBSD is `u32` rather than the expected `u64` like on other OSes.


.builds/freebsd.yml
```diff
@@ -0,0 +1,18 @@
+image: freebsd/latest
+sources:
+  - https://github.com/sunfishcode/cap-std
+tasks:
+  - setup: |
+      curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y --default-toolchain=none --profile=minimal
+      source $HOME/.cargo/env
```
sunfishcode, line 7: Should `$HOME` be whitespace-protected by doublequotes? Or can we use `~` instead? (here and below)

cap-primitives/src/yanix/common/fs/metadata_ext.rs
```diff
@@ -101,7 +101,7 @@ impl std::os::unix::fs::MetadataExt for MetadataExt {
 
     #[inline]
     fn ino(&self) -> u64 {
-        self.ino
+        self.ino.into()
```
sunfishcode, line 104: Is this `into()` needed? `ino` here is our own `u64` struct field.

kubkon:
OK, I think this one is ready to go!

PR #23, opened 2020-07-08
Remove sys submodules in cap-std and cap-async-std

kubkon:
Majority of host-specific impls are now done in `cap-primitives`,
and the remaining host-specific bits are few and far between that
they can be included in the main `fs` module in `cap-std`/`cap-async-std`.

WASI is an exception here, but given that we don't have full support for it yet,
I figure we can refactor further when WASI arrives.


cap-async-std/src/fs/dir.rs
```diff
@@ -66,13 +72,55 @@ impl Dir {
...
+    }
+
+    #[cfg(windows)]
+    fn _open_file_with(
+        file: &std::fs::File,
+        path: &Path,
+        options: &OpenOptions,
+    ) -> io::Result<File> {
+        open(file, path.as_ref(), options)
+            .map(|f| unsafe { File::from_raw_handle(f.into_raw_handle()) })
```
sunfishcode, line 95: Could we add a `File::from_sync_file` and factor out the fd/handle parts so that Unix and Windows can otherwise share an `_open_file_with`?

src/fs/dir.rs
```diff
@@ -65,13 +70,36 @@ impl Dir {
         path: P,
         options: &OpenOptions,
     ) -> io::Result<File> {
-        self.sys.open_file_with(path.as_ref(), options)
+        self._open_file_with(path.as_ref(), options)
+    }
+
+    #[cfg(not(target_os = "wasi"))]
+    fn _open_file_with(&self, path: &Path, options: &OpenOptions) -> io::Result<File> {
+        open(&self.std_file, path.as_ref(), options).map(File::from_std)
```
sunfishcode, line 78: This `as_ref()` shouldn't be needed.

cap-async-std/src/fs/dir.rs
```diff
@@ -66,13 +72,55 @@ impl Dir {
...
+        Self::_open_file_with(&file, path.as_ref(), options)
+    }
+
+    #[cfg(any(unix, target_os = "fuchsia"))]
+    fn _open_file_with(
+        file: &std::fs::File,
+        path: &Path,
+        options: &OpenOptions,
+    ) -> io::Result<File> {
+        open(file, path.as_ref(), options).map(|f| unsafe { File::from_raw_fd(f.into_raw_fd()) })
```
sunfishcode, line 85: Ditto about `.as_ref()`.

cap-async-std/src/fs/dir.rs
```diff
@@ -1,19 +1,27 @@
...
+    } else if #[cfg(windows)] {
+        use cap_primitives::fs::{symlink_dir, symlink_file};
+        use async_std::os::windows::io::{AsRawHandle, FromRawHandle, IntoRawHandle, RawHandle};
+    } else if #[cfg(target_os = "wasi")] {
+        use async_std::os::wasi::{
+            fs::OpenOptionsExt,
+            io::{AsRawFd, IntoRawFd},
+        };
+    }
+}
```
sunfishcode, line 23: I think we can avoid `cfg_if` here and in the other places that do this, because the conditions are already disjoint, so we don't need `else`. To avoid repeating the condition for multiple use statements we can merge them:
```rust
use {
   crate::os::unix::net::UnixDataGram,
   cap_primitives::fs::canonicalize,
}
so that one `cfg` shouldsuffiice.

kubkon: Oh niiice! I love this! Lemme give it a whirl!

cap-async-std/src/fs/dir.rs
kubkon: Actually, it turns out there's a `From` conversion provided by `async_std`, so we can use `File::from_std` in combination with `From` :-)

PR #25, opened 2020-07-09
Draft out Windows support

kubkon:
This PR drafts out Windows support. The majority of required low-level primitives are left as stubs, however, I did make sure it builds at least (currently, only on Nightly due to the way `libstd` traits are defined in Rust).

Also, while here, I've gone ahead and copied over @alexcrichton's "install-rust" from [bytecodealliance/wasmtime] repo to streamline toolchain installation in GA.

[bytecodealliance/wasmtime]: https://github.com/bytecodealliance/wasmtime


cap-async-std/src/fs_utf8/mod.rs
```diff
@@ -39,18 +39,20 @@ fn from_utf8<P: AsRef<str>>(path: P) -> std::io::Result<std::path::PathBuf> {
...
+        let bytes = string.as_cstr().to_bytes();
         OsStr::from_bytes(bytes).to_owned()
     };
 
     #[cfg(windows)]
     let path = {
-        use std::{ffi::OsString, os::windows::prelude::*};
-        OsString::from_wide(bytes)
+        use std::{ffi::OsString, os::windows::ffi::OsStringExt};
+        let utf8 = string.as_cstr().to_string_lossy(); // This is OK since we went through arf_strings first.
```
sunfishcode, line 53: This is coming out the "back end" of arf_strings, so `string` isn't guaranteed to be valid UTF-8; it's an arbitrary NUL-terminated byte sequence. I think this whole area is something we're going to have to figure out within the arf-strings library itself. Probably we'll need a `WindowsString` counterpart to `PosixString`.

For now, to be safe, instead of `to_string_lossy()`, how about `to_str()?`. That way we won't risk accidental collisions due to lossiness.

cap-async-std/src/fs_utf8/mod.rs
```diff
@@ -68,7 +70,14 @@ fn to_utf8<P: AsRef<std::path::Path>>(path: P) -> std::io::Result<String> {
...
     #[cfg(windows)]
-    let cstr = { panic!("windows not yet implemented") };
+    let cstr = {
+        use std::{ffi::CString, os::windows::ffi::OsStrExt};
+        let utf16: Vec<_> = osstr.encode_wide().collect();
+        let str = String::from_utf16(&utf16).map_err(|_| {
+            std::io::Error::new(std::io::ErrorKind::InvalidData, "invalid path string")
+        })?;
+        CString::new(str)?
+    };
```
sunfishcode, line 80: This code looks good. Long term, I think arf-strings will also need a `WasiString::from_maybe_nonutf16` or so, but we can figure that out separately.

cap-primitives/src/winx/fs/metadata_ext.rs
```diff
@@ -0,0 +1,94 @@
...
+        }
+    }
+
+    /// Determine if `self` and `other` refer to the same inode on the same device.
+    #[cfg(debug_assertions)]
+    pub(crate) fn is_same_file(&self, other: &Self) -> bool {
+        // From [MSDN]:
+        // The identifier (low and high parts) and the volume serial number uniquely identify a file on a single computer.
+        // To determine whether two open handles represent the same file, combine the identifier and the volume serial
+        // number for each file and compare them.
```
sunfishcode, line 40: Could you reflow this paragraph to fit in 100 columns?

cap-primitives/src/winx/fs/mkdir_unchecked.rs
```diff
@@ -0,0 +1,6 @@
+use std::{fs, io, path::Path};
+
+/// *Unsandboxed* function similar to `mkdir`, but which does not perform sandboxing.
+pub(crate) fn mkdir_unchecked(start: &fs::File, path: &Path) -> io::Result<()> {
+    unimplemented!("mkdir_unchecked")
```
sunfishcode, line 5: These can use `todo!` :-)

cap-async-std/src/fs/file.rs
```diff
@@ -210,28 +210,34 @@ impl fmt::Debug for File {
...
+    let fd = fd.as_raw_fd();
+    b.field("fd", &fd);
+    if let Some((read, write)) = unsafe { get_mode(fd) } {
+        b.field("read", &read).field("write", &write);
     }
 }
+
+#[cfg(windows)]
+fn fmt_debug_file(fd: &impl AsRawHandle, b: &mut fmt::DebugStruct) {
+    // TODO fill in the blanks
```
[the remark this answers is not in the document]
kubkon: Good idea, thanks!

cap-async-std/src/fs_utf8/mod.rs
kubkon: Ah ok. I wasn’t sure about this one. I’ll be happy to add the missing bits to arf-strings to support utf16. In the meantime, I’ll tweak this to use lossless `to_str`.

cap-primitives/src/winx/fs/metadata_ext.rs
kubkon: Sure thing!

Cargo.toml
```diff
@@ -30,6 +30,7 @@ libc = "0.2.71"
 
 [target.'cfg(windows)'.dependencies]
 winx = "0.18.0"
+winapi = { version = "^0.3", features = ["winbase"] }
```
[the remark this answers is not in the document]
kubkon: Oh sorry, this must a remnant from `wasi-common` where I copied the stuff from.

kubkon:
Hmm, I think I screwed up the CI cause tests are not triggered. Lemme fix that before me merge this in.

kubkon:
> Hmm, I think I screwed up the CI cause tests are not triggered. Lemme fix that before we merge this in.

OK, that should be it.

PR #26, opened 2020-07-09
Ensure tests build on Windows

kubkon:
Also, ensure `TempDir` can actually be created on Windows, and add `mkdir_unchecked`.

EDIT: ~~**Depends on #25**~~


tests/fs.rs
```diff
@@ -812,8 +812,8 @@ fn copy_file_returns_metadata_len() {
     let in_path = "in.txt";
     let out_path = "out.txt";
     check!(check!(tmp.create(&in_path)).write(b"lettuce"));
-    #[cfg(windows)]
-    check!(check!(tmp.create(tmp.join("in.txt:bunny"))).write(b"carrot"));
+    // #[cfg(windows)]
+    // check!(check!(tmp.create(tmp.join("in.txt:bunny"))).write(b"carrot"));
```
sunfishcode, line 816: Could you add a comment mentioning why this is commented out, with a TODO if appropriate?

kubkon:
@sunfishcode I've now re-enabled the commented out assertion, but also went ahead and added `get_path` implementation targeting Windows. I've also reshuffled the helpers on Windows (`concatenate` and friends into `cap-prim..::winx::fs::get_path::*` module). Finally, since the calls to for instance `mkdir_unchecked` are meant to be well unchecked, instead of returning an error if the `path` argument is in fact absolute, on Windows we now simply return the `path` unmodified to essentially emulate POSIX.

PR #28, opened 2020-07-10
More consistency checking

sunfishcode:
Add more consistency checks to canonicalize, link, and convert `MaybeOwnedFile` into a struct which performs additional checking.

### DIFF
--- a/cap-primitives/src/std/fs/canonicalize.rs
+++ b/cap-primitives/src/std/fs/canonicalize.rs
@@ -3,7 +3,7 @@
 
 use crate::fs::canonicalize_impl;
 #[cfg(debug_assertions)]
-use crate::fs::{is_same_file, open, OpenOptions};
+use crate::fs::{get_path, is_same_file, open, OpenOptions};
 use std::{
     fs, io,
     path::{Path, PathBuf},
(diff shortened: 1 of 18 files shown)


cap-primitives/src/std/fs/canonicalize.rs
```diff
@@ -28,7 +29,36 @@ pub fn canonicalize(start: &fs::File, path: &Path) -> io::Result<PathBuf> {
...
+            assert_eq!(
+                check_abspath,
+                result_abspath,
+                "incorrect canonicalization: start='{:?}' path='{}' result='{}'",
+                start,
+                path.display(),
+                canonical_path.display()
+            );
+            assert!(
+                result_abspath.starts_with(start_abspath),
```
kubkon, line 56: I think this check might fail on Windows since `start_abspath` will *not* contain the extended path prefix `\\?\`, while the result of `fs::canonicalize` always does:

> On Windows, this converts the path to use extended length path syntax...

Anyhow, I reckon this is OK to leave as-is for now since we don't have a working implementation on Windows yet anyhow, but if you could leave a comment to that effect as a reminder, that'd be great!

sunfishcode: Commented!

PR #29, opened 2020-07-13
Rename some things

sunfishcode:
`cap_primitives::std` and `cap_primitives::yanix::common::fs` were workarounds for me not being super familiar with `mod.rs` logic. 

`resolve_symlink_at` doesn't make it clear that it's just doing a single `readlink` on a single path component.

`winx::fs::get_path` avoids duplication at the callsite.


sunfishcode:
Rebased and conflicts fixed, but now Windows is mysteriously failing.

kubkon:
> Rebased and conflicts fixed, but now Windows is mysteriously failing.

Lemme provide a quick fix since I've got a Windows handy.

sunfishcode:
I think I may have just figured it out. I was missing a `mod get_path;` after the `get_path` rename. Let's see if that fixes it.

kubkon:
> I think I may have just figured it out. I was missing a `mod get_path;` after the `get_path` rename. Let's see if that fixes it.

Oh and actually, since it's a compile error, we could just as easily cross-verified on any OS. All it takes is to install the matching target and then run `cargo check --target x86_64-pc-windows-msvc`. I always keep forgetting that it's a possibility. ;-)

kubkon:
@sunfishcode OK, f9590d7 should hopefully fix it!

sunfishcode:
Indeed it does, thanks!

PR #30, opened 2020-07-14
Add a cap-primitives fuzzer and fix numerous fuzz bugs.

sunfishcode:
Split checking logic out of the main cap-primitives API functions into
separate `check_*` functions, to keep the main code path readable.

Treat empty paths as errors, as POSIX does.

Make `open_parent` handle `..` automatically by using `.` as the
basename rather than making all callers deal with it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,2 @@
-/target
+target
 Cargo.lock
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -47,4 +47,5 @@ fs_utf8 = ["arf-strings"]
 members = [
   "cap-primitives",
   "cap-async-std",
+  "fuzz",
 ]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -47,5 +47,6 @@ fs_utf8 = ["arf-strings"]
 members = [
   "cap-primitives",
   "cap-async-std",
-  "fuzz",
+  # Work around https://github.com/rust-lang/cargo/issues/8338.
+  #"fuzz",
 ]
--- a/cap-primitives/Cargo.toml
+++ b/cap-primitives/Cargo.toml
@@ -13,6 +13,7 @@ readme = "README.md"
 [dependencies]
 ipnet = "2.3.0"
 cfg-if = "0.1.9"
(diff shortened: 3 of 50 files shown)


cap-primitives/src/std/fs/canonicalize.rs
```diff
@@ -19,12 +19,19 @@ pub fn canonicalize(start: &fs::File, path: &Path) -> io::Result<PathBuf> {
...
+
+#[cfg(debug_assertions)]
+fn check_canonicalize(start: &fs::File, path: &Path, result: &io::Result<PathBuf>) {
+    if let Ok(canonical_path) = result {
         let path_result = open(start, path, OpenOptions::new().read(true));
         let canonical_result = open(start, canonical_path, OpenOptions::new().read(true));
         match (path_result, canonical_result) {
             (Ok(path_file), Ok(canonical_file)) => {
-                assert!(is_same_file(&path_file, &canonical_file)?)
+                assert!(is_same_file(&path_file, &canonical_file).unwrap())
```
kubkon, line 34: Actually, I think it would be better from the point of view of bug tracing to panic with a custom message here? I.e., `expect` instead of `unwrap`.

cap-primitives/src/std/fs/canonicalize_manually.rs
```diff
@@ -1,22 +1,36 @@
...
+    follow: FollowSymlinks,
+) -> io::Result<PathBuf> {
     let mut symlink_count = 0;
     let mut canonical_path = PathBuf::new();
 
     if let Err(e) = open_manually(
         start,
         path,
-        OpenOptions::new().read(true),
+        follow.options(OpenOptions::new().read(true)),
```
kubkon, line 33: Hmm, I'm not sure about this one. Would it perhaps be more readable if we had a method similar to `OpenOptions::nofollow` but instead of a `bool`, it would take `FollowSymlinks` as an argument?

sunfishcode: Good idea! I've now done that, and it cleans up several areas of code.

kubkon:
OK, I'm going to merge this as is, and in terms of CI coverage, we can iterate in-tree.

PR #39, opened 2020-07-16
Ensure fuzz_targets are built in the CI

kubkon:
Also, add a short `README` on fuzzing.


kubkon:
> This looks great! Should we also update CONTRIBUTING.md's fuzzing section point to this README.md instead of having its own less-complete info?

Yes! Doing it just now!

PR #42, opened 2020-07-17
Convert source files from CRLF to LF endings.

sunfishcode:
This just keeps files in the repo consistent.


kubkon:
Awesome, thanks! I'll need to make sure to configure my editor to use LF by default on Windows VM!

PR #43, opened 2020-07-17
Implement more features for Linux etc.

sunfishcode:
This implements `remove_dir`, `read_dir`, `remove_dir_all`.


sunfishcode:
I just realized that `MaybeOwnedFile` could implement `Deref` instead of `AsRef<File>`, which makes it a little tidier.

PR #56, opened 2020-07-22
Use OS-specific error codes for checking for symlinks with O_NOFOLLOW.

sunfishcode:
NetBSD uses yet another error code, so add code to check for OS-specific
errno codes.

### DIFF
--- a/cap-primitives/src/yanix/fs/open_unchecked.rs
+++ b/cap-primitives/src/yanix/fs/open_unchecked.rs
@@ -30,7 +30,19 @@ pub(crate) fn open_unchecked(
         }
     };
     match err.raw_os_error() {
-        Some(libc::ELOOP) | Some(libc::EMLINK) => Err(OpenUncheckedError::Symlink(err)),
+        // `ELOOP` is the POSIX standard and most widely used error code to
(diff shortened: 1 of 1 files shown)


cap-primitives/src/yanix/fs/open_unchecked.rs
```diff
@@ -30,7 +30,19 @@ pub(crate) fn open_unchecked(
...
+        #[cfg(not(any(target_os = "freebsd", target_os = "dragonfly", target_os = "netbsd")))]
+        Some(libc::ELOOP) => Err(OpenUncheckedError::Symlink(err)),
+
+        // FreeBSD and similar (but not Darwin) use `EMLINK`.
+        #[cfg(any(target_os = "freebsd", target_os = "dragonfly"))]
+        Some(libc::EMLINK) => Err(OpenUncheckedError::Symlink(err)),
+
+        // NetBSD uses `EFTYPE`.
+        #[cfg(any(target_os = "netbsd"))]
+        Some(libc::EFTYPE) => Err(OpenUncheckedError::Symlink(err)),
```
kubkon, line 44: I need to figure out a CI for this target somehow.

PR #61, opened 2020-07-22
Add more missing Windows functionality

kubkon:
This commit adds `readlink_unchecked` and implements various missing
bits in `dir_options` module. While here, it seems that `readlink_one`
could be shared between hosts, and hence, it moves that out into
a module directly under host-independent `fs` module.

It also cleans up the use of `winapi` crate. Since in most places
(except for `cap-primitives`), `winapi` was used solely to extract
the flag `FILE_FLAG_BACKUP_SEMANTICS`, it is possible and better
to use the corresponding `winx::file::Flags::FILE_FLAG_BACKUP_SEMANTICS`
instead.


src/fs/dir.rs
```diff
@@ -111,12 +111,12 @@ impl Dir {
...
     fn _open_dir(&self, path: &Path) -> io::Result<Self> {
         use std::os::windows::fs::OpenOptionsExt;
-        use winapi::um::winbase::FILE_FLAG_BACKUP_SEMANTICS;
+        use winx::file::Flags;
         self.open_with(
             path,
             OpenOptions::new()
                 .read(true)
-                .attributes(FILE_FLAG_BACKUP_SEMANTICS),
+                .attributes(Flags::FILE_FLAG_BACKUP_SEMANTICS.bits()),
```
sunfishcode, line 119: Can this use `dir_options()` instead of a manual `OpenOptions`? It seems like that might also eliminate the need for `cap-std` and `cap-async-std` to depend directly on `winx`.

cap-primitives/src/fs/readlink_one.rs
```diff
@@ -28,7 +29,7 @@ pub(crate) fn readlink_one(
     );
 
     if *symlink_count == MAX_SYMLINK_EXPANSIONS {
```
sunfishcode, line 31: I like the idea of making `readlink_one` platform-independent, but I have one concern: do you know if Windows also has a concept of a `MAX_SYMLINK_EXPANSIONS` in the way that Unix-type platforms do? We'd ideally match Windows path lookup behavior here, complete with symlink recursion detection.

kubkon: Good catch! Windows indeed does have a limit on reparse points and that limit currently is set at 63. I've now made `MAX_SYMLINK_EXPANSIONS` constant host-dependent, which effectively means we can still keep `readlink_one` host-independent, and we pull in from host-specific modules only the error and the constant.

src/fs/dir.rs
kubkon: Excellent idea, thanks!

cap-primitives/src/fs/mod.rs
```diff
@@ -51,8 +53,10 @@ pub(crate) use unlink_via_parent::*;
 cfg_if::cfg_if! {
     if #[cfg(any(unix, target_os = "fuchsia"))] {
         pub(crate) use super::yanix::fs::*;
+        pub use super::yanix::fs::dir_options::dir_options;
```
sunfishcode, line 56: Ah, I forgot that `dir_options` was in `cap-primitives` while `_open_dir` is in `cap-std`. It feels like `dir_options` and the fact that we use `open` to open directories is an implementation detail of `cap-primitives`. Rather than having `cap-primitives` publish `dir_options` here, could we add an `open_dir` function to `cap-primitives`?

Cargo.toml
```diff
@@ -30,7 +30,6 @@ libc = "0.2.72"
 
 [target.'cfg(windows)'.dependencies]
 winx = "0.19.0"
```
sunfishcode, line 32: Can we remove the `witx` dependency here?

cap-async-std/Cargo.toml
```diff
@@ -31,7 +31,6 @@ libc = "0.2.72"
 
 [target.'cfg(windows)'.dependencies]
 winx = "0.19.0"
```
sunfishcode, line 33: And here? :-)

cap-primitives/src/yanix/fs/mod.rs
```diff
@@ -56,11 +55,12 @@ pub(crate) use open_options_ext::*;
...
-pub(crate) use readlink_one::*;
 pub(crate) use readlink_unchecked::*;
 pub(crate) use remove_dir_all_impl::*;
 pub(crate) use rename_unchecked::*;
 pub(crate) use rmdir_unchecked::*;
 pub(crate) use stat_unchecked::*;
 pub(crate) use symlink_unchecked::*;
 pub(crate) use unlink_unchecked::*;
+
+pub(crate) const MAX_SYMLINK_EXPANSIONS: u8 = 40;
```
sunfishcode, line 66: We can cite https://man7.org/linux/man-pages/man7/path_resolution.7.html as the source for `40` here. This may need to be customized for non-Linux platforms, but I think we can figure that out later.

Cargo.toml
kubkon: So we could, but given that we'll most likely need it similarly to the way we need `yanix` for debug printing I thought I was going to leave it in, rather than remove it now just to add it back in. Unless we won't need `winx` for that at all... OK, let's remove it for now, and add it in the future if need be.

cap-primitives/src/fs/mod.rs
kubkon: Hmm, sounds fair. Lemme see what I can come up with!

Cargo.toml
sunfishcode: Ah, then perhaps we should figure out how to move some part of the debug printing into `cap-primitives` so that we don't need `yanix` here either :-). It'll still be a transitive dependency, but the separation will help us avoid accidentally adding too many implementation details to the outer `cap-std` crate.

kubkon: Excellent idea! If you don't mind though, leave it for a subsequent PR?

cap-primitives/src/fs/mod.rs
kubkon: OK, I've now created a new module in `cap-primitives::fs::open_dir` which is essentially a wrapper around `open` with `dir_options` pre-specified as `OpenOptions` and reused that inside `cap-std` and `cap-async-std`.

PR #64, opened 2020-07-23
Make the `O_PATH` emulation logic handle other errors from `readlinkat`.

sunfishcode:


kubkon:
Interesting, it seems we're failing on linux with kernel >=5.6.

sunfishcode:
Huh, I was going by the documentation there, and thought `readlinkat` would return `EINVAL` if the file wasn't a symlink. Looks like it actually returns `ENOENT` in this case. Fixed now.

PR #67, opened 2020-07-23
Re-add winx as a dev-dependency for cap-std.

sunfishcode:
This fixes cargo test with windows targets.

While here, make the libc dev-dependency conditional.


sunfishcode:
This was fixed in #65.

PR #71, opened 2020-07-24
Create `cap-tempfile` and `cap-directories` crates.

sunfishcode:
Like `tempfile` and `directories`, but capability-oriented; you get `Dir`s
instead of `Path`s.

### DIFF
--- a/cap-tempfile/src/lib.rs
+++ b/cap-tempfile/src/lib.rs
@@ -1,7 +1,7 @@
 //! Capability-oriented temporary directories.
 
 use cap_std::fs::Dir;
-use std::{fs, io, ops::Deref};
+use std::{io, ops::Deref};
 
 /// A directory in a filesystem that is automatically deleted when it goes out of scope.
 ///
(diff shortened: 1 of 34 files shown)


cap-tempfile/src/lib.rs
```diff
@@ -0,0 +1,59 @@
...
+
+impl TempDir {
+    /// Attempts to make a temporary directory inside of `env::temp_dir()`.
+    ///
+    /// This corresponds to [`tempfile::TempDir::new`].
+    ///
+    /// [`tempfile::TempDir::new`]: https://docs.rs/tempfile/latest/tempfile/struct.TempDir.html#method.new
+    pub fn new() -> io::Result<Self> {
+        let inner = tempfile::TempDir::new()?;
+        let dir = Dir::from_std_file(fs::File::open(inner.path())?);
```
kubkon, line 28: Since under-the-hood, `std::fs::File::open(&path)` translates into

```rust
std::fs::OpenOptions::new().read(true).open(&path)
```

this will fail on Windows. What we need here is the logic from `tests::sys_common`:

```rust
use std::os::windows::fs::OpenOptionsExt;
use winx::file::Flags;

let dir = Dir::from_std_file(fs::OpenOptions::new()
    .read(true)
    .write(true)
    .attributes(Flags::FILE_FLAG_BACKUP_SEMANTICS.bits())
    .open(&path)?
);
```

sunfishcode: Ok. I expect we'll have more crates that do this, in addition to cap-directories and cap-tempfile, so I created a utility routine in cap-primitives to do this, which is now exposed in cap-std as `Dir::open_ambient_dir`. It's `unsafe`, not because of memory safety or even file-descriptor safety, but under the interpretation of `unsafe` as "doesn't uphold the invariants of the API", since it's not sandboxed. What do you think?

kubkon: Sounds perfect, thanks!

PR #76, opened 2020-07-27
Add a simple key-value CLI example for `cap-std`.

sunfishcode:


sunfishcode:
While building this example, I realized that `cap-directories`' `ProjectDirs` represents an ambient authority. This is something we'll likely iterate on, but for now, to represent this accurately, `ProjectDirs::new` is marked `unsafe`.

kubkon:
Could you rebase please?

PR #94, opened 2020-08-04
Add `Dir::remove_open_dir`, `cap_tempdir::tempdir_in`, and friends.

sunfishcode:
This adds functions to remove the directory represented by a `Dir`
directly, and uses it to implement `cap_tempdir::tempdir_in`.

Note that it's generally not possible to perform `remove_open_dir`
atomically on popular OS's, so document it accordingly.


sunfishcode:
After a bit of iterating, this is now at the point where what it needs is `read_dir` for Windows.

PR #115, opened 2020-10-16
Add another test for hardlinks to symlinks.

sunfishcode:


sunfishcode:
I've now filed https://github.com/rust-lang/rust/pull/78026 about the difference in the `std::fs` version of this test.

PR #137, opened 2021-01-15
Minimal optimization for cap-async-std.

sunfishcode:
Use `spawn_blocking` to run important blocking tests off the main
thread.


sunfishcode:
This PR depends on https://github.com/async-rs/async-std/pull/937 making it to a published async-std.

sunfishcode:
Two functions to revisit here: `DirEntry::metadata` and `DirEntry::file_type`.

PR #149, opened 2021-02-02
Add `remove_file_or_symlink` to DirExt and DirExtUtf8

pchickey:


cap-fs-ext/src/dir_ext.rs
```diff
@@ -157,6 +169,18 @@ pub trait DirExtUtf8 {
     fn open_dir_nofollow<P: AsRef<str>>(&self, path: P) -> io::Result<Self>
     where
         Self: Sized;
+
+    /// Removes a file or symlink from a filesystem.
+    ///
+    /// Removal of symlinks has different behavior under Windows - if a symlink points
+    /// to a directory, it cannot be removed with the remove_file operation. This
+    /// method will remove files and all sy mlinks
```
sunfishcode, line 177: Here and above, typo: "sy mlinks". Also, please add a period at the end of the sentence.

cap-fs-ext/src/dir_ext.rs
```diff
@@ -157,6 +169,18 @@ pub trait DirExtUtf8 {
...
+    ///
+    /// This corresponds to [`std::fs::remove_file`], but only accesses paths
+    /// relative to `self`.
```
sunfishcode, line 180: Here and above, the "This corresponds to" sentences aren't needed here since these don't directly correspond to things in std.

cap-fs-ext/src/dir_ext.rs
```diff
@@ -238,6 +262,24 @@ impl DirExt for cap_std::fs::Dir {
...
+    #[cfg(windows)]
+    #[inline]
+    fn remove_file_or_symlink<P: AsRef<Path>>(&self, path: P) -> io::Result<()> {
+        match self.remove_file(path.as_ref()) {
+            Ok(()) => Ok(()),
+            Err(e) => match self.remove_dir(path.as_ref()) {
+                Ok(()) => Ok(()),
+                Err(_) => Err(e),
+            },
+        }
```
sunfishcode, line 281: The unfortunate part here is that we can't avoid removing an empty directory here. But I think I convinced myself that's ok; it's only an empty directory after all. But, we should mention that in the documentation comment.

cap-fs-ext/src/dir_ext.rs
pchickey: e6cb016

cap-fs-ext/src/dir_ext.rs
pchickey: e6cb016

cap-fs-ext/src/dir_ext.rs
pchickey: Good point! e6cb016

PR #155, opened 2021-02-04
DirExt::remove_file_or_symlink no longer deletes empty directory

pchickey:
(except in the case of a race, which is pretty contrived)


sunfishcode (approved):
Looks good! As we discussed, we can experiment with ways to avoid removing the directory separately.

PR #156, opened 2021-02-16
Add support for files whose canonical paths use UNC prefixes.

sunfishcode:
On Windows, files on network shares have a canonical path which starts
with the UNC "\\?\" prefix. With these prefixes, the OS no longer
handles "." paths and trailing slashes, so handle them manually.

Fixes #2650.


sunfishcode:
Looking at the test failures, this needs some more work before it's ready.

sunfishcode:
Windows has some functions which perform canonicalization. I've now submitted https://github.com/retep998/winapi-rs/pull/992 to add bindings to these in winapi, so we can experiment with using those instead of trying to do the canonicalization manually.

sunfishcode:
Another option is to use [windows-rs](https://github.com/microsoft/windows-rs). That seems like the next thing to try here.

sunfishcode:
A better path forward for this issue is to change cap-std to [use `NtCreateFile`](https://github.com/bytecodealliance/cap-std/issues/226), which is available as a windows-sys API, so it's stable to use. That way we won't have to do path concatenation and deal with all of these subtle canonicalization and UNC details.

PR #158, opened 2021-02-17
 Make `remove_file_or_symlink_dir` non-racy.

sunfishcode:
On Windows, `remove_file` doesn't remove the file until the last handle
is closed, so hold the handle open while we check the symlink's type to
avoid race conditions.


pchickey (approved):
Of course. This is much better.

PR #164, opened 2021-03-08
Build fix for s390x platform

uweigand:
On s390x, PROC_SUPER_MAGIC (like all superblock magic numbers)
must have type libc::c_uint instead of libc::__fsword_t.

This is done correctly in the libc crate, but not in the local copy
of PROC_SUPER_MAGIC in cap-primitives/src/posish/linux/fs/procfs.rs,
causing a build failure on s390x.

Fix by duplicating the libc logic here as well.


sunfishcode (approved):
Thanks!

uweigand:
Just a quick question following up on this: are you planning to create a new version of the crate anytime soon?  I think this is needed so I can update the dependency in wasmtime to fix the build failure there.

sunfishcode:
I've now released 0.13.9 with this patch in it.

uweigand:
Works for me, thanks!

PR #166, opened 2021-04-14
Use the `AmbientAuthority` type instead of `unsafe` to indicate ambient authority

### LINKED ISSUE #141: Is `unsafe` the best way to indicate ambient authority?
cap-std crates currently use `unsafe` to indicate function which utilize ambient authorities, for example [`Dir::open_ambient_dir`](https://docs.rs/cap-std/0.11.0/cap_std/fs/struct.Dir.html#method.open_ambient_dir). As discussed [here](
https://github.com/bytecodealliance/cap-std/issues/129#issuecomment-752143974), this follows the precedent of Rust's standard library in making [`File::from_raw_fd`](https://doc.rust-lang.org/stable/std/os/unix/io/trait.FromRawFd.html#tymethod.from_raw_fd) be unsafe.

However, when porting code to cap-std, such as [this Web server example](https://github.com/sunfishcode/tide-naive-static-files) and looking at it from the perspective of a user of the API incrementally adopting it, the use of `unsafe` has felt somewhat uncomfortable. So, what's the best approach here?

Some possible options include:
 - Write a linting system. See also [this issue](https://github.com/bytecodealliance/cap-std/issues/130). This is a fair amount of work, and would be great for users that want to scan large codebases for ambient authorities, however it'd be less convenient for incremental users, such as the Web server example mentioned above.
 - Put `unsafe` functions in a separate module, or even a separate crate.
 - Continue to use `unsafe`.

sunfishcode:
Use the ambient-authority crate instead of `unsafe`.
    
Use the newly-created ambient-authority crate to indicate functions
which have ambient authority, rather than using `unsafe`.

Fixes #141.

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,9 +91,8 @@ There currently are three main ways:
  - Use the [`cap-directories`] crate to create `Dir`s for config, cache and
    other data directories.
  - Use the [`cap-tempfile`] crate to create `Dir`s for temporary directories.
- - Use the `unsafe` [`Dir::open_ambient_dir`] to open a plain path. This
-   function is not sandboxed, and may open any file the host process has
-   access to.
+ - Use [`Dir::open_ambient_dir`] to open a plain path. This function is not
+   sandboxed, and may open any file the host process has access to.
 
 ## Examples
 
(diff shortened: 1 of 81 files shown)


sunfishcode:
This works reasonably well, and does eliminate the last `unsafe` code from cap-primitives, which is nice. I want to do some experimenting with this from a user perspective before committing to it though.

PR #173, opened 2021-06-13
cap-tempfile: dev-depend on winapi only on Windows

olivierlemasle:
`cap-tempfile` should have a dev-dependency on `winapi` only on Windows.

Only usage on `winapi` in `cap-tempfile`: https://github.com/bytecodealliance/cap-std/blob/1ec2258d7b03b4a4f64108cf4bf59003f0aa7ab0/cap-tempfile/src/lib.rs#L214-L218


sunfishcode:
Good catch, thanks!

PR #174, opened 2021-06-14
Port to io-lifetimes, add optional linux-raw-sys support

sunfishcode:
This eliminates all direct dependencies from cap-std on libc, so that we can optionally enable posish's linux-raw-sys support. And, it adds io-lifetimes support and trait implementations.

For the moment, it depends on https://github.com/bytecodealliance/posish/pull/10/


sunfishcode:
bytecodealliance/posish#10 has landed, and other fixes are in.

PR #194, opened 2021-11-04
Update to rustix 0.26.0 and Rust 2021 edition

sunfishcode:
With Rust 2021 edition, `TryInto` is now in the prelude, and we can do `for thing in [...]` now!


sunfishcode:
At present, we have users using Rust 1.48, so I'll revert the Rust 2021 change and resubmit this.

PR #197, opened 2021-11-10
On Android, use an explicit version check to detect `openat2`.

sunfishcode:
On Android, seccomp is configured to kill processes that execute
unsupported system calls, rather than returning `ENOSYS`, which
breaks the auto-detection code. So on Android, use an explicit
version check, using `uname`, to test whether the running kernel
supports `openat2`, before trying to call it.


hrydgard:
Works!

sunfishcode:
Thanks!

PR #207, opened 2021-11-26
reduce scope of ambient_authority in fs example

### LINKED ISSUE #206: reduce scope of ambient_authority in examples?
The `ambient_authority` function is in scope for the whole file:

https://github.com/bytecodealliance/cap-std/blob/c936cbb518a2e070571d4bff1cdc85aaf7d5608f/examples/std_fs_misc.rs#L4

Is there some way to leave it out of scope of everything but the main function? I should re-read how rust modules work.

Maybe main() would only call `open_ambient_dir` and then pass `cwd` to a function that does the rest?

dckc:
please excuse the completely untested suggestion
_p.s. ah... the ci bots can check this._

fixes #206

### DIFF
--- a/examples/std_fs_misc.rs
+++ b/examples/std_fs_misc.rs
@@ -103,4 +103,3 @@ fn main() {
 
     file_stuff(&mut cwd);
 }
-
--- a/examples/std_fs_misc.rs
+++ b/examples/std_fs_misc.rs
@@ -1,7 +1,6 @@
 // Copied from https://doc.rust-lang.org/rust-by-example/std_misc/fs.html and
 // adapted to use this crate instead.
 
-use cap_std::ambient_authority;
 use cap_std::fs::{Dir, OpenOptions};
 use std::io;
 use std::io::prelude::*;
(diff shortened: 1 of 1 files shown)


sunfishcode:
Could you add some comments to the example describing the specific things that users should learn from it? I myself am a little unclear on what the lesson here should be.

sunfishcode:
I'm inclined to decline this specific change. In practice, what matters are actual calls to `ambient_authority`. Minimizing the scope of the `ambient_authority` identifier doesn't prevent users from calling it; it just discourages users from calling it frequently and makes those calls stand out more. However, `ambient_authority` is already a fairly verbose identifier which is inconvenient to call frequently, and it already stands out when it's called.

Reducing the scope of `ambient_authority` can even make code harder to understand and audit. If code really needs `ambient_authority()` in multiple places, it's usually best for it to simply call it in multiple places, rather than taking on extra complexity to be able to use it from multiple places while having only a single callsite.

Does this make sense?

dckc:
fair enough

PR #212, opened 2022-01-07
fs::dir: Add a `try_reopen_dirfd()` method

cgwalters:
In some cases when working with an existing C/C++ codebase already
using `openat()` or other crates such as the `openat` crate, one
may have an existing file descriptor and want to create a `cap-std`
view of the same directory.

Add an API which reopens from a source FD.

### DIFF
--- a/cap-std/src/fs/dir.rs
+++ b/cap-std/src/fs/dir.rs
@@ -609,6 +609,20 @@ impl Dir {
         let _ = ambient_authority;
         fs::create_dir_all(path)
     }
+
+    /// Construct a new instance of `Self` from existing directory file descriptor.
+    ///
+    /// This can be useful when interacting with other libraries and or C/C++ code
+    /// which has invoked `openat(..., O_DIRECTORY)` external to this crate.
+    #[cfg(not(windows))]
+    pub fn reopen_dir(fd: BorrowedFd) -> io::Result<Self> {
+        use io_lifetimes::AsFilelike;
+        cap_primitives::fs::open_dir(
+            &fd.as_filelike_view::<std::fs::File>(),
+            std::path::Component::CurDir.as_ref(),
+        )
+        .map(Self::from_std_file)
+    }
 }
 
 #[cfg(not(windows))]
--- a/tests/fs_additional.rs
+++ b/tests/fs_additional.rs
@@ -847,3 +847,13 @@ fn maybe_dir() {
     // Opening directories works on all platforms with `maybe_dir`.
     check!(tmpdir.open_with("dir", OpenOptions::new().read(true).maybe_dir(true)));
 }
+
+#[test]
+#[cfg(not(windows))]
+fn reopen_fd() {
+    use rustix::fd::AsFd;
+    let tmpdir = tmpdir();
+    check!(tmpdir.create_dir("subdir"));
+    let tmpdir2 = check!(cap_std::fs::Dir::reopen_dir(tmpdir.as_fd()));
+    assert!(tmpdir2.exists("subdir"));
+}


cgwalters:
xref https://github.com/coreos/rpm-ostree/pull/3311

cgwalters:
OK right, doesn't build on Windows.  So far I've successfully spent most of my 20+ years of programming not having to think much about Windows :smile:  Do we need an extension trait here instead?  For now I just added a `cfg(not(windows))`.

sunfishcode:
We may revisit this later, but I think `cfg(not(windows))` is fine in this situation. I'll do a proper review hopefully in a few days here.

cgwalters:
I wonder if this use case may come up elsewhere, and the "unsafe borrow then manually drop" is obviously error prone.  What do you think about this on top?  Or hmm, maybe even a default method of the [AsFd trait](https://docs.rs/io-lifetimes/0.4.4/io_lifetimes/trait.AsFd.html)?

```
diff --git a/cap-primitives/src/fs/mod.rs b/cap-primitives/src/fs/mod.rs
index 7085368..6cf9570 100644
--- a/cap-primitives/src/fs/mod.rs
+++ b/cap-primitives/src/fs/mod.rs
@@ -85,7 +85,7 @@ pub use remove_dir_all::remove_dir_all;
 pub use remove_file::remove_file;
 pub use remove_open_dir::{remove_open_dir, remove_open_dir_all};
 pub use rename::rename;
-pub use reopen::reopen;
+pub use reopen::{map_borrowed_fd_as_std, reopen};
 #[cfg(not(target_os = "wasi"))]
 pub use set_permissions::set_permissions;
 pub use set_times::{set_times, set_times_nofollow};
diff --git a/cap-primitives/src/fs/reopen.rs b/cap-primitives/src/fs/reopen.rs
index b19bbcd..7afeb96 100644
--- a/cap-primitives/src/fs/reopen.rs
+++ b/cap-primitives/src/fs/reopen.rs
@@ -1,5 +1,7 @@
 //! Re-open a `fs::File` to produce an independent handle.
 
+use io_lifetimes::BorrowedFd;
+
 use crate::fs::{is_file_read_write, is_same_file, reopen_impl, OpenOptions};
 use std::{fs, io};
 
@@ -34,3 +36,21 @@ pub fn reopen(file: &fs::File, options: &OpenOptions) -> io::Result<fs::File> {
 
     Ok(new)
 }
+
+/// Given a borrowed file descriptor, view it as a borrowed [`std::fs::File`] for the duration of a closure.
+///
+/// A primary use is to create a copy of the file descriptor using an API that expects a std File.
+#[cfg(not(windows))]
+#[allow(unsafe_code)]
+pub fn map_borrowed_fd_as_std<F, T>(fd: BorrowedFd, f: F) -> io::Result<T>
+where
+    F: Fn(&std::fs::File) -> io::Result<T>,
+{
+    use rustix::fd::{AsRawFd, FromRawFd, IntoRawFd};
+    // Temporary view
+    let view = unsafe { std::fs::File::from_raw_fd(fd.as_raw_fd()) };
+    let r = f(&view);
+    // Now, drop the std wrapper for the fd - we don't want to close it!
+    let _ = view.into_raw_fd();
+    r
+}
diff --git a/cap-std/src/fs/dir.rs b/cap-std/src/fs/dir.rs
index c473343..d3f6d29 100644
--- a/cap-std/src/fs/dir.rs
+++ b/cap-std/src/fs/dir.rs
@@ -616,13 +616,12 @@ impl Dir {
     /// which has invoked `openat(..., O_DIRECTORY)` external to this crate.
     #[cfg(not(windows))]
     pub fn try_reopen_dirfd(fd: BorrowedFd) -> io::Result<Self> {
-        // Temporarily view this as a std File
-        let src = unsafe { std::fs::File::from_raw_fd(fd.as_raw_fd()) };
-        // Then convert into a cap-std file
-        let r = cap_primitives::fs::open_dir(&src, std::path::Component::CurDir.as_ref())?;
-        // Now, drop the std wrapper for the fd - we don't want to close it!
-        let _ = src.into_raw_fd();
-        Ok(Self::from_std_file(r))
+        cap_primitives::fs::map_borrowed_fd_as_std(fd, |fd| {
+            Ok(Self::from_std_file(cap_primitives::fs::open_dir(
+                &fd,
+                std::path::Component::CurDir.as_ref(),
+            )?))
+        })
     }
 }
 
```

sunfishcode:
io-lifetimes has a [`as_filelike_view`](https://docs.rs/io-lifetimes/latest/io_lifetimes/trait.AsFilelike.html#tymethod.as_filelike_view) utility which internally does the same thing. Using that, the code could look something like this:
```rust
cap_primitives::fs::open_dir(fd.as_filelike_view::<fs::File>(), std::path::Component::CurDir.as_ref())?;
```

cgwalters:
Ah yes, that feeling when you're playing chess and then you realize the other player has been 2 steps ahead of you for a while

cap-std/src/fs/dir.rs
```diff
@@ -609,6 +609,20 @@ impl Dir {
         let _ = ambient_authority;
         fs::create_dir_all(path)
     }
+
+    /// Construct a new instance of `Self` from existing directory file descriptor.
+    ///
+    /// This can be useful when interacting with other libraries and or C/C++ code
+    /// which has invoked `openat(..., O_DIRECTORY)` external to this crate.
+    #[cfg(not(windows))]
+    pub fn try_reopen_dirfd(fd: BorrowedFd) -> io::Result<Self> {
```
sunfishcode, line 618: Could we name this `reopen_dir`?
 - I think we can omit the `try_` because other "open" functions in cap-std and std don't have `try_` in their names, as it's commonly understood that opening something can fail.
 - Change "dirfd" to "dir" because I think we'll be able to generalize this to Windows in the future, with basically the same API, but without calling it "fd", so dropping the "fd" now will reduce churn later.

sunfishcode (commented):
This looks great, just one more further detail:

cap-std/src/fs/dir.rs
cgwalters: Makes sense.  I think the reason I had `try_` in there is because I had in my head in the background that I also wanted to add a `try_exists` to cap-std (xref https://internals.rust-lang.org/t/the-api-of-path-exists-encourages-broken-code/13817)

And also related to that originally I was thinking of this as a "clone" operation, but it's a `try_clone()`.

But I agree `reopen_dir()` is clear and to the point, so fixed!

PR #215, opened 2022-01-10
Document that the openat crate doesn't follow symlinks.

sunfishcode:
Add a note to the related crates section of the top-level README
mentioning that the openat crate doesn't follow symlinks, while cap-std
does (and sandboxes them).


cgwalters (approved):

sunfishcode:
Thanks for the review!

PR #218, opened 2022-01-14
Add features corresponding to Rust 1.58's new filesystem features.

sunfishcode:
Add `Metadata::is_symlink` and `File::options`, corresponding to the new
std features in [Rust 1.58].

This also updates tests/fs.rs and tests/fs_utf8.rs to the latest changes
from upstream Rust.

[Rust 1.58]: https://blog.rust-lang.org/2022/01/13/Rust-1.58.0.html

### DIFF
--- a/tests/fs_utf8.rs
+++ b/tests/fs_utf8.rs
@@ -1504,11 +1504,15 @@ fn create_dir_long_paths() {
     path.extend(iter::repeat(OsStr::new("a")).take(PATH_LEN - utf16_len));
 
     // This should succeed.
-    tmpdir.create_dir(&path).unwrap();
+    tmpdir
+        .create_dir(&PathBuf::from_path_buf(path.clone().into()).unwrap())
+        .unwrap();
 
     // This will fail if the path isn't converted to verbatim.
     path.push("a");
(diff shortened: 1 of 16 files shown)


tests/fs_utf8.rs
```diff
@@ -1504,11 +1504,15 @@ fn create_dir_long_paths() {
     path.extend(iter::repeat(OsStr::new("a")).take(PATH_LEN - utf16_len));
 
     // This should succeed.
-    tmpdir.create_dir(&path).unwrap();
+    tmpdir
+        .create_dir(&PathBuf::from_path_buf(path.clone().into()).unwrap())
```
cgwalters, line 1508: (I try to heavily use camino where it makes sense)

cgwalters (approved):
LGTM

tests/fs_utf8.rs
sunfishcode: This particular diff is awkward because it's ported from a std test which eagerly converts the path into an `OsString`, and I attempted to port it to fs_utf8 with minimal changes. In normal usage, once you have a `Utf8PathBuf`, you'd just keep it as such, and not convert it to `OsString` :-).

PR #221, opened 2022-01-20
Expose cap_std::fs_utf8 on docs.rs

sunshowers:
I was trying to use `fs_utf8` for some of my code, and docs.rs didn't
have any documentation for it.

I'm not sure if this is desired yet, but I wanted to put this PR up to
solicit feedback.

Also enable doc_cfg + doc_auto_cfg so that docs.rs marks fs_utf8 (and
net) as being conditional.

image

### DIFF
--- a/cap-std/Cargo.toml
+++ b/cap-std/Cargo.toml
@@ -12,9 +12,12 @@ categories = ["filesystem", "network-programming"]
 repository = "https://github.com/bytecodealliance/cap-std"
 edition = "2018"
 
+[package.metadata.docs.rs]
+rustdoc-args = ["--all-features", "--cfg=doc_cfg"]
+
 [dependencies]
 arf-strings = { version = "0.6.3", optional = true }
-cap-primitives = { path = "../cap-primitives", version = "^0.22.2-alpha.0"}
+cap-primitives = { path = "../cap-primitives", version = "^0.22.2-alpha.0" }
 ipnet = "2.3.0"
 io-extras = "0.12.0"
 io-lifetimes = { version = "0.4.0", default-features = false }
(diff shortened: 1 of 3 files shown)


sunfishcode:
This looks like a good idea! Would it make sense to also add a comment somewhere prominent saying something like "to use the fs_utf8 module, enable the fs_utf8 cargo feature"?

sunshowers:
Done, thanks!

sunfishcode:
Thanks!

cgwalters:
This seems to have broken the build on docs.rs: https://docs.rs/crate/cap-std/0.23.0/builds/496356

cgwalters:
OK I am not a rustdoc guru, but it looks to me like this is closer to correct:

```
diff --git a/cap-std/Cargo.toml b/cap-std/Cargo.toml
index fe1eb07..88bd279 100644
--- a/cap-std/Cargo.toml
+++ b/cap-std/Cargo.toml
@@ -13,7 +13,7 @@ repository = "https://github.com/bytecodealliance/cap-std"
 edition = "2018"
 
 [package.metadata.docs.rs]
-rustdoc-args = ["--all-features", "--cfg=doc_cfg"]
+all-features = true
 
 [dependencies]
 arf-strings = { version = "0.6.3", optional = true }
```

What I don't understand though is the sorcery involved in the docs.rs build that is distinct from `cargo +nightly` doc locally.

sunshowers:
Ahh sorry about that! all-features = true is correct but we do need --cfg=doc_cfg in the rustdoc-args.

cgwalters:
OK, I took a swing at this in https://github.com/bytecodealliance/cap-std/pull/225
It'd be nice if docs.rs more clearly printed the cargo invocation it's using without being entangled with the docker bits.

sunfishcode:
cap-std 0.23.1 is now published with the fix, and the docs.rs build is now fixed, and it has the `fs_utf8` docs. Thanks!

sunshowers:
Thanks again!

PR #225, opened 2022-01-22
cap-std: Fix the build on docs.rs

cgwalters:
Followup to https://github.com/bytecodealliance/cap-std/pull/221


sunshowers:
looks good!

PR #229, opened 2022-01-30
README.md: Link to cap-std-ext

cgwalters:
I am hopeful we can drain most of the functionality here into
cap-std itself or cap-fs-ext, but for now let's link to it so others
can find it more easily.

Most of this would be nice to have in Rust `std` too; that's a whole
other TODO item.  Perhaps best tackled once `std` itself depends on
rustix.


sunfishcode:
Looks good! If you'd like, you could add a mention of rpm-ostree in the Examples section in the README.md too.

PR #231, opened 2022-01-31
Update to io-lifetimes 0.5.1.

sunfishcode:
This updates to io-lifetimes 0.5.1, in cap-std itself and in its
dependencies.

io-lifetimes is pretty stable at this point so hopefully these
kinds of semver-breaking releases won't happen frequently. In this
case, io-lifetimes needed a version bump due to the `HandleOrNull`
change, which only affects Windows, and only affects low-level ffi
users of io-lifetimes.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,10 +25,10 @@ rand = "0.8.1"
 tempfile = "3.1.0"
 camino = "1.0.5"
 libc = "0.2.100"
-io-lifetimes = "0.4.4"
+io-lifetimes = "0.5.1"
 
 [target.'cfg(not(windows))'.dev-dependencies]
-rustix = "0.32.0"
+rustix = "0.33.0"
 
 [target.'cfg(windows)'.dev-dependencies]
 # nt_version uses internal Windows APIs, however we're only using it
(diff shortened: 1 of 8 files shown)


Cargo.toml
```diff
@@ -25,10 +25,10 @@ rand = "0.8.1"
 tempfile = "3.1.0"
 camino = "1.0.5"
 libc = "0.2.100"
-io-lifetimes = "0.4.4"
+io-lifetimes = "0.5.1"
 
 [target.'cfg(not(windows))'.dev-dependencies]
-rustix = "0.32.0"
+rustix = "0.33.0"
```
cgwalters, line 31: Ahh nice and this is what I need for https://github.com/coreos/rpm-ostree/pull/3374#issuecomment-1025818269

(Seems like the commit message/PR title should mention rustix too, as I doubt I'm the only one who will be using *both* cap-std and rustix and hence wanting a single version of both)

cgwalters (approved):

Cargo.toml
sunfishcode: I edited the title to mention rustix via the squash-and-merge UI.

sunfishcode:
This is now released as 0.24.0.

cgwalters:
Thanks, I did https://github.com/cgwalters/cap-std-ext/pull/3

PR #232, opened 2022-02-03
Add documentation about `symlink`'s argument order.

sunfishcode:
The order of the arguments to the `symlink` function could reasonably go
either of two different ways. And it appears cap-std picked one way and
the openat crate [picked the other way]. Both ways have a reason why they
make sense, and both ways also have the potential for user surprise.

cap-std's overarching reason for its choice is to follow std as closely
as possible. std's `symlink`, following POSIX `symlink` and `ln -s`,
puts the link name second. So a user converting `std` code to `cap-std`
code would just need to add a `Dir` as the `self` argument, without
needing to reorder anything. I think that still makes sense, but I'm
open to feedback here.

One thing that makes sense no matter which way is chosen is to document it,
so this patch adds documentation comments. And, it renames the arguments
from `src` and `dst` to `original` and `link` to match the naming used in
the `symlink` function in std.

[picked the other way]: https://docs.rs/openat/0.1.21/openat/struct.Dir.html#method.symlink

### DIFF
--- a/cap-async-std/src/fs/dir.rs
+++ b/cap-async-std/src/fs/dir.rs
@@ -535,27 +535,47 @@ impl Dir {
 
     /// Creates a new symbolic link on a filesystem.
     ///
+    /// The `original` argument provides the target of the symlink. The `link`
+    /// argument provides the name of the created symlink.
+    ///
+    /// Despite the argument ordering, `original` is not resolved relative to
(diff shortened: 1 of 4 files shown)


cgwalters (approved):
Right, perhaps you saw me fall into this trap :hole: a few times :smile: 

> And, it renames the arguments
from src and dst to original and link to match the naming used in
the symlink function in std.

Yes!  This alone is a big help!

PR #233, opened 2022-02-03
Add a mention of `unix_fd` as a similar crate.

sunfishcode:
`unix_fd` has a `Chroot` type which is also a directory type that
performs sandboxing, however similar to `pathrs`, it also resolves
absolute paths as being relative to the root.

### DIFF
--- a/README.md
+++ b/README.md
@@ -205,13 +205,15 @@ sandbox, while `openat` doesn't support following symlinks.
 
 `cap-std` has some similar functionality to [`pathrs`] in that it also
 explicitly verifies that `/proc` has actual `procfs` mounted on it and nothing
-mounted on top, and it can also use `openat2`. However, `cap-std` uses
-`RESOLVE_BENEATH`-style resolution where absolute paths are considered errors,
-while `pathrs` uses `RESOLVE_IN_ROOT` where absolute paths are interpreted as
-references to the base file descriptor. And overall, `cap-std` seeks to provide
-a portable `std`-like API which supports Windows in addition to Unix-like
-platforms, while `pathrs` provides a lower-level API that exposes more of the
-underlying `openat2` options and only supports Linux.
+mounted on top, and it can also use `openat2`. And it has some similar
+functionality to [`unix_fd`]. However, `cap-std` uses `RESOLVE_BENEATH`-style
+resolution where absolute paths are considered errors, while `pathrs` and
+`unix_fd` use `RESOLVE_IN_ROOT`-style resolution, where absolute paths are
+interpreted as references to the base file descriptor. And overall, `cap-std`
+seeks to provide a portable `std`-like API which supports Windows in addition
+to Unix-like platforms, while `pathrs` provides a lower-level API that exposes
+more of the underlying `openat2` options and only supports Linux, and `unix_fd`
+is specific to Unix-like platforms.
 
 [`obnth`] is a new crate which appears to be very similar to `cap_std::fs`.
 It's not mature yet, and it doesn't support Windows. It does support
@@ -223,3 +225,4 @@ and `RESOLVE_IN_ROOT`, including emulation when `openat2` isn't available.
 [`pathrs`]: https://crates.io/crates/pathrs
 [`obnth`]: https://crates.io/crates/obnth
 [`camino`]: https://crates.io/crates/camino
+[`unix_fd`]: https://crates.io/crates/unix_fd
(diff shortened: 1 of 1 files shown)


cgwalters (approved):
Only tangentially related but honestly I've been kind of wanting `RESOLVE_IN_ROOT` by default.  May be worth calling out explicitly why you chose `RESOLVE_BENEATH` for cap-std?

sunfishcode:
> Only tangentially related but honestly I've been kind of wanting `RESOLVE_IN_ROOT` by default. May be worth calling out explicitly why you chose `RESOLVE_BENEATH` for cap-std?

I've now added some words about `RESOLVE_BENEATH`. However, I'd be interested in learning more about your use case. Could you say more about how you use `Dir`?

cgwalters:
> I've now added some words about RESOLVE_BENEATH.

:+1: 

> However, I'd be interested in learning more about your use case. Could you say more about how you use Dir?

In (rpm-)ostree, like a lot of container tooling we are running in one root filesystem, and operate on (or construct) a new fully distinct one.  I think this use case is exactly why `RESOLVE_IN_ROOT` was created.  While those of us making OSes and containers I think generally use relative symlinks, there are still sometimes absolute ones and I want to at least not fail on those.

Relatedly, it's just a bit tedious to have to use relative paths for e.g. `/etc/resolv.conf` which is really a global resource - in that rootfs.

How about a new `Dir::open_rooted` API that opts-in to `RESOLVE_IN_ROOT`?

sunfishcode:
> Relatedly, it's just a bit tedious to have to use relative paths for e.g. `etc/resolv.conf` which is really a global resource - in that rootfs.
> 
> How about a new `Dir::open_rooted` API that opts-in to `RESOLVE_IN_ROOT`?

For opening paths like "etc/resolv.conf" and others, `open_rooted` might be a good feature for cap-std-ext. That seems like it should be straightforward.

Doing it that way wouldn't handle absolute-path symlinks, but one option would be to be opinionated ;-). Arguably, symlinks to absolute paths are already a hazard in an environment where you may have logical filesystem roots living at non-root positions in the process' filesystem namespace. While it's technically possible to handle those symlinks, it's arguably better to surface them as errors early.

If you really do need absolute-path symlink support, my concern would be that `open_rooted` seems like it wouldn't be sufficient; I imaginee you'd want all functions that take a path to have a `_rooted` form (because even non-following functions like `mkdir` still follow symlinks in parent-directory components of their paths). That would add a lot of clutter, and it would require users to be careful to type `_rooted` at every call, which seems error-prone. So if we're going to do something like that, we should think about ways to organize the API.

cgwalters:
> For opening paths like "etc/resolv.conf" and others, open_rooted might be a good feature for cap-std-ext. That seems like it should be straightforward.

Yeah, I may do that.  

> If you really do need absolute-path symlink support, my concern would be that open_rooted seems like it wouldn't be sufficient; I imaginee you'd want all functions that take a path to have a _rooted form (because even non-following functions like mkdir still follow symlinks in parent-directory components of their paths).

Couldn't we do something like

```
diff --git a/cap-std/src/fs/dir.rs b/cap-std/src/fs/dir.rs
index 5b39247..5b6c632 100644
--- a/cap-std/src/fs/dir.rs
+++ b/cap-std/src/fs/dir.rs
@@ -43,6 +43,7 @@ use {
 /// [functions in `std::fs`]: https://doc.rust-lang.org/std/fs/index.html#functions
 pub struct Dir {
     std_file: fs::File,
+    is_rooted: bool,
 }
 ```

And then dispatch on that to tell cap-primitives/rustix to use `RESOLVE_IN_ROOT` instead?

cgwalters:
To be very clear, this is completely a sidebar discussion, which I find interesting and useful, but should definitely not block merging this PR!

sunfishcode:
My concern with a `is_rooted` field is that it changes the contract for existing cap-std users. With such a field, when someone passes you a `Dir`, it might be a rooted `Dir` or it might not be. What would you think about a new type, say `Root`, which would have a similar API to `Dir`, but would do `RESOLVE_IN_ROOT` for all paths?

Separately, I'm a little concerned about the complexity of passing an `is_rooted` flag through all the cap-primitives filesystem code. It's more than just passing the right flag to `openat2`; there's also the fallback code, which we need even on new Linux versions due to the openat2`/`EAGAIN` issue. That said, this isn't a show-stopper.

cgwalters:
>  What would you think about a new type, say Root, which would have a similar API to Dir, but would do RESOLVE_IN_ROOT for all paths?

Sounds reasonable to me! 
It's also tempting honestly to forcibly rewrite all symlinks to be relative in our build process.  The people doing container tools would probably be more hesitant to do that as it's possible it could break something in apps.

Anyways, this is definitely not an immediate need.  I'm just thinking through the ergonomics.  Thanks for the discussion!

PR #234, opened 2022-02-08
dir: Add `try_exists()`

cgwalters:
I want to use this for the same reason as
https://github.com/rust-lang/rust/pull/81822
was added to std.  While it's currently unstable, I can't imagine
it would change.

### DIFF
--- a/cap-std/src/fs/dir.rs
+++ b/cap-std/src/fs/dir.rs
@@ -584,6 +584,24 @@ impl Dir {
         self.metadata(path).is_ok()
     }
 
+    /// Returns `true` if the path points at an existing entity.
+    ///
+    /// This corresponds to [`std::fs::try_exists`], but only
+    /// accesses paths relative to `self`.
+    ///
+    /// # API correspondence with `std`
+    ///
+    /// This API is not yet stable in `std`, but is likely to be.  For more
+    /// information, see the [tracker issue](https://github.com/rust-lang/rust/issues/83186).
+    #[inline]
+    pub fn try_exists<P: AsRef<Path>>(&self, path: P) -> io::Result<bool> {
+        match self.metadata(path) {
+            Ok(_) => Ok(true),
+            Err(error) if error.kind() == io::ErrorKind::NotFound => Ok(false),
+            Err(error) => Err(error),
+        }
+    }
+
     /// Returns `true` if the path exists on disk and is pointing at a regular
     /// file.
     ///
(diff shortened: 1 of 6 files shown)


cap-std/src/fs/dir.rs
```diff
@@ -584,6 +584,19 @@ impl Dir {
         self.metadata(path).is_ok()
     }
 
+    /// Returns `true` if the path points at an existing entity.
+    ///
+    /// This corresponds to [`std::path::Path::try_exists`], but only
```
sunfishcode, line 589: Here and in the other instances of this, I'd prefer to reference `std::fs::try_exists` here, because most of the `Path` forms are just aliases for the `std::fs` forms, and I find it conceptually unfortunate that an otherwise I/O-free `Path` type has these I/O-performing convenience aliases attached to it.

tests/dir-ext.rs
```diff
@@ -11,7 +11,7 @@ fn remove_file() {
     let file = tempdir.create("file").expect("create file to delete");
     drop(file);
     tempdir.remove_file_or_symlink("file").expect("delete file");
-    assert!(!tempdir.exists("file"), "deletion worked");
+    assert!(!tempdir.try_exists("file").unwrap(), "deletion worked");
```
sunfishcode, line 14: Can we leave the `exists` test here, and have both `exists` and `try_exists` tests here?

tests/fs_additional.rs
```diff
@@ -191,6 +191,12 @@ fn optionally_recursive_mkdir() {
     assert!(tmpdir.is_dir(dir));
 }
 
+#[test]
+fn try_exists() {
+    let tmpdir = tmpdir();
+    assert_eq!(tmpdir.try_exists("somefile").unwrap(), false);
```
sunfishcode, line 197: It'd be interesting to add a test for "somedir/somefile" too.

sunfishcode (commented):
cap-std usually puts features corresponding to unstable features in std behind a config option, however in this case, I think it's reasonable to just provide these. As you say, it's unlikely that these will change before stabilization, and they're useful.

Please add comments to the cap-std versions mentioning that the `std` version is not yet stabilized though, with a link to the tracking issue. And please add comments to the cap-async-std versions mentioning that cap-async-std doesn't have these operations yet.

cgwalters:
OK, comments should be addressed!

sunfishcode:
This is now released in 0.24.2.

PR #235, opened 2022-02-09
Fix compilation errors on Rust nightly.

sunfishcode:
Fix compilation errors that started appering on Rust nightly. This
appears to just need an explicit dereference where one was not
previously needed.

An example of the error message is:

```
error[E0308]: mismatched types
  --> cap-primitives/src/fs/via_parent/rename.rs:22:58
   |
22 |     let (old_dir, old_basename) = open_parent(old_start, &old_path)?;
   |                                                          ^^^^^^^^^ expected struct `Path`, found opaque type
   |
  ::: cap-primitives/src/rustix/fs/dir_utils.rs:67:48
   |
67 | pub(crate) fn strip_dir_suffix(path: &Path) -> impl Deref<Target = Path> + '_ {
   |                                                ------------------------------ the found opaque type
   |
   = note:   expected struct `Path`
           found opaque type `impl Deref<Target = Path>`
```

### DIFF
--- a/cap-primitives/src/fs/via_parent/create_dir.rs
+++ b/cap-primitives/src/fs/via_parent/create_dir.rs
@@ -13,7 +13,7 @@ pub(crate) fn create_dir(start: &fs::File, path: &Path, options: &DirOptions) ->
     // slashes.
     let path = strip_dir_suffix(path);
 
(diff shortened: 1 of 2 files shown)


sunfishcode:
Merging this trivial change so that we can get an update out to fix build breakage.

PR #237, opened 2022-03-16
Update to io-lifetimes 0.7.0.

sunfishcode:
io-lifetimes needed a semver bump due to `BorrowedFd::borrow_raw_fd`
being renamed to `BorrowedFd::borrow_raw`, to match the corresponding change
in std. cap-std doesn't use this function itself, but since it's a semver
bump, we need to bump it in cap-std so that it uses the new trait versions.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,7 +35,7 @@ rustix = { version = "0.35.6", features = ["fs"] }
 # nt_version uses internal Windows APIs, however we're only using it
 # for testing here.
 nt_version = "0.1.3"
-winapi = "0.3.9"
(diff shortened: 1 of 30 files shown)


sunfishcode:
io-lifetimes will need another semver bump for https://github.com/sunfishcode/io-lifetimes/pull/27 and any other changes requested by the std maintainers, so it's likely that cap-std here will wait on this PR until that's ready.

cgwalters (approved):

sunfishcode:
Rustix 0.35 is now released, and this PR is now updated to use it.

PR #238, opened 2022-03-16
Prefer to use `O_PATH` in `set_permissions` on Linux.

sunfishcode:
This patch makes two changes:

 - It removes the code that tries `O_PATH` + `fchmod`, as `fchmod` is
   [documented] to fail on `O_PATH` file descriptors, which now seems
   unlikely to ever change.

 - It moves the call to `set_permissions_through_proc_self_fd` to the
   top so that it's the first strategy tried. This function uses
   `O_PATH` + `fchmodat` on /proc/self/fd/{}, which may fail due to
   missing support for `O_PATH` or procfs not being mounted, but if it
   does succeed, it's the best option because opening files with
   `O_PATH` tells Linux we don't need `read`/`write`/etc. access, which
   may allow it to succeed in more cases, and may have fewer side
   effects when opening device files.

[documented]: https://man7.org/linux/man-pages/man2/open.2.html#DESCRIPTION

### DIFF
--- a/cap-primitives/src/rustix/linux/fs/set_permissions_impl.rs
+++ b/cap-primitives/src/rustix/linux/fs/set_permissions_impl.rs
@@ -1,10 +1,8 @@
 use super::procfs::set_permissions_through_proc_self_fd;
 use crate::fs::{errors, open, OpenOptions, Permissions};
-use rustix::fs::{fchmod, Mode, OFlags, RawMode};
-use std::os::unix::fs::{OpenOptionsExt, PermissionsExt};
+use rustix::fs::{fchmod, Mode, RawMode};
+use std::os::unix::fs::PermissionsExt;
 use std::path::Path;
-use std::sync::atomic::AtomicBool;
-use std::sync::atomic::Ordering::Relaxed;
 use std::{fs, io};
 
 pub(crate) fn set_permissions_impl(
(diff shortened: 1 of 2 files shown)


cap-primitives/src/rustix/linux/fs/set_permissions_impl.rs
```diff
@@ -1,49 +1,27 @@
...
-            }
-        }
+    // First try using `O_PATH` and doing a chmod on `/proc/self/fd/{}`
+    // (`fchmod` doesn't work on `O_PATH` file descriptors).
+    //
+    // This may fail, due to older Linux versions not supporting `O_PATH`, or
+    // due to procfs being unavailable, but if it does work, go with it, as
+    // `O_PATH` tells Linux that we don't actually need to read or write the
+    // file, which may avoid side effects associated with opening device files.
+    let result = set_permissions_through_proc_self_fd(start, path, std_perm.clone());
```
cgwalters, line 22: Not that I care much, but in theory we could cache the failure here just like above?

(But eh, anyone running such old kernels has bigger problems IMO)

cgwalters (approved):
It is a bit confusing to me that indirecting through `/proc` would make this work.

cap-primitives/src/rustix/linux/fs/set_permissions_impl.rs
sunfishcode: Yeah, we could, but also yeah. It's useful to make things work, to the extent we can, but `O_PATH` goes back to 2.6.39, so it's unlikely this level of optimization is important here.

sunfishcode:
This is now released in 0.24.2.

PR #239, opened 2022-03-25
tempfile: Add support for `TempFile`

cgwalters:
The `cap-tempfile` crate ironically only supports temporary
directories right now.

This adapts code I wrote as part of https://github.com/coreos/cap-std-ext/
to add support for temporary files that can also be atomically
linked into place.

### DIFF
--- a/cap-tempfile/src/tempfile.rs
+++ b/cap-tempfile/src/tempfile.rs
@@ -1,0 +1,255 @@
+//! Temporary files.
+
+use cap_std::fs::{Dir, File};
+use std::ffi::OsStr;
+use std::io::{self, Write};
+use std::io::{Read, Seek};
+
+/// A file in a directory that is by default deleted when it goes out
+/// of scope, but may also be written persistently.
+///
+/// This corresponds most closely to [`tempfile::NamedTempFile`]; however,
+/// there are some important differences, so read the below carefully
+/// to understand how to port existing code.
+///
+/// # Name-able, but not necessarily named
+///
+/// By default, the file does not necessarily have an name until the file is written
+/// persistently.
+///
+/// On some operating systems like Linux, it is possible to create anonymous
+/// temporary files that can still be written to disk persistently via `O_TMPFILE`.
+/// The advantage of this is that if the process (or operating system) crashes
+/// while the file is being written, the temporary space will be automatically cleaned up.
+/// For this reason, there is no API to retrieve the name, for either case.
+///
+/// To more closely match the semantics of [`tempfile::tempfile`], use [`crate::TempFile::new_anonymous`].
+///
+/// # File permissions
+///
+/// Unlike the tempfile crate, the default [`TempFile::new`] will use the same permissions as [`File::create_new`] in
+/// the Rust standard library.  On Unix for example, this is `0o666` modified by `umask`.
+/// The rationale for this is to make it more ergonomic and natural to use this API to atomically create new files
+/// and replace existing ones.
+///
+/// [`tempfile::NamedTempFile`]: https://docs.rs/tempfile/latest/tempfile/struct.NamedTempFile.html
+/// [`File::create_new`]: https://doc.rust-lang.org/std/fs/struct.OpenOptions.html#method.create_new
+pub struct TempFile<'d> {
+    dir: &'d Dir,
+    fd: File,
+    name: Option<String>,
+}
+
+#[cfg(any(target_os = "android", target_os = "linux"))]
+fn new_tempfile_linux(d: &Dir) -> io::Result<Option<File>> {
+    use rustix::fd::FromFd;
+    use rustix::fs::{Mode, OFlags};
+    // openat's API uses WRONLY.  There may be use cases for reading too, so let's support it.
+    let oflags = OFlags::CLOEXEC | OFlags::TMPFILE | OFlags::RDWR;
+    // We default to 0o666, same as main rust when creating new files; this will be modified by
+    // umask: https://github.com/rust-lang/rust/blob/44628f7273052d0bb8e8218518dacab210e1fe0d/library/std/src/sys/unix/fs.rs#L762
+    let mode = Mode::from_raw_mode(0o666);
+    // Happy path - Linux with O_TMPFILE
+    match rustix::fs::openat(d, ".", oflags, mode) {
+        Ok(r) => return Ok(Some(File::from_fd(r.into()))),
+        // See https://github.com/Stebalien/tempfile/blob/1a40687e06eb656044e3d2dffa1379f04b3ef3fd/src/file/imp/unix.rs#L81
+        Err(rustix::io::Error::OPNOTSUPP | rustix::io::Error::ISDIR | rustix::io::Error::NOENT) => {
+            Ok(None)
+        }
+        Err(e) => {
+            return Err(e.into());
+        }
+    }
+}
+
+/// Assign a random name to a currently anonymous O_TMPFILE descriptor.
+#[cfg(any(target_os = "android", target_os = "linux"))]
+fn generate_name_in(subdir: &Dir, f: &File) -> io::Result<String> {
+    use rustix::fd::AsFd;
+    use rustix::fs::AtFlags;
+    let procself_fd = rustix::io::proc_self_fd()?;
+    let fdnum = rustix::path::DecInt::from_fd(&f.as_fd());
+    let fdnum = fdnum.as_c_str();
+    super::retry_with_name_ignoring(io::ErrorKind::AlreadyExists, |name| {
+        rustix::fs::linkat(&procself_fd, fdnum, subdir, name, AtFlags::SYMLINK_FOLLOW)
+            .map_err(Into::into)
+    })
+    .map(|(_, name)| name)
+}
+
+/// Create a new temporary file in the target directory, which may or may not have a (randomly generated) name at this point.
+fn new_tempfile(d: &Dir) -> io::Result<(File, Option<String>)> {
+    // On Linux, try O_TMPFILE
+    #[cfg(any(target_os = "android", target_os = "linux"))]
+    if let Some(f) = new_tempfile_linux(d)? {
+        return Ok((f, None));
+    }
+    // Otherwise, fall back to just creating a randomly named file.
+    let mut opts = cap_std::fs::OpenOptions::new();
+    opts.read(true);
+    opts.write(true);
+    opts.create_new(true);
+    super::retry_with_name_ignoring(io::ErrorKind::AlreadyExists, |name| {
+        d.open_with(name, &opts)
+    })
+    .map(|(f, name)| (f, Some(name)))
+}
+
+impl<'d> TempFile<'d> {
+    /// Crate a new temporary file in the provided directory.
+    pub fn new(dir: &'d Dir) -> io::Result<Self> {
+        let (fd, name) = new_tempfile(dir)?;
+        Ok(Self { dir, fd, name })
+    }
+
+    /// Crate a new temporary file in the provided directory that will not have a
+    /// name.  This corresponds to [`tempfile::tempfile_in`].
+    ///
+    /// [`tempfile::tempfile_in`]: https://docs.rs/tempfile/latest/tempfile/fn.tempfile_in.html
+    pub fn new_anonymous(dir: &'d Dir) -> io::Result<File> {
+        let (fd, name) = new_tempfile(dir)?;
+        if let Some(name) = name {
+            dir.remove_file(name)?;
+        }
+        Ok(fd)
+    }
+
+    /// Get a reference to the underlying file.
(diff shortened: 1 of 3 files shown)


cap-tempfile/src/tempfile.rs
```diff
@@ -0,0 +1,223 @@
...
+        d.open_with(name, &opts)
+    })
+    .map(|(f, name)| (f, Some(name)))
+}
+
+#[cfg(unix)]
+fn default_permissions() -> Permissions {
+    use std::os::unix::prelude::PermissionsExt;
+    Permissions::from_mode(0o600)
+}
```
cgwalters, line 81: (I don't yet know how to do this right on Windows)

Also...*ideally* on Unix we'd do something like calculate the default permissions that would apply after `umask` is invoked.  Defaulting to `0600` is clearly *safe* but many users will need to change it.

cgwalters:
Side note...I have kind of lost track of how many times I have implemented Linux O_TMPFILE stuff.  A long time ago I wrote a C implementation of this in ostree, then cleaned it up and factored it out into https://gitlab.gnome.org/GNOME/libglnx where I also did fd-relative i.e. C `openat()` stuff (but still not capability/`openat2` based).

Then...Rust came along, I eventually decided to build on [openat](https://docs.rs/openat/latest/openat/) and reimplement it *again* in [openat-ext](https://docs.rs/openat-ext/latest/openat_ext/).

So this is at least the 4th time, but I am pretty sure I also did some stuff around this in [glib](https://gitlab.gnome.org/GNOME/glib) but at this point that's getting to be a distant memory.

But anyways, I for one am looking forward to going into the second half of my career mostly writing Rust code, and also hopefully not writing basic code to manipulate temporary files again...

cgwalters:
OK I belatedly realized that having the API pass in `Permissions` was really just a workaround for a [long ago fixed glibc bug](https://sourceware.org/bugzilla/show_bug.cgi?id=17523).  I think we can relatively safely rely on that being fixed - if some user shows up bitten by it, we could have rustix do the direct syscall path, or perhaps try to detect it and fall back to not using `O_TMPFILE`.

(The latter path will want the "compute current process umask" code that lives in the test suite for now)

sunfishcode (commented):
> OK I belatedly realized that having the API pass in `Permissions` was really just a workaround for a [long ago fixed glibc bug](https://sourceware.org/bugzilla/show_bug.cgi?id=17523). I think we can relatively safely rely on that being fixed - if some user shows up bitten by it, we could have rustix do the direct syscall path, or perhaps try to detect it and fall back to not using `O_TMPFILE`.

According to the linked bug report, the bug is present in glibc 2.20, however Rust itself is only now [updating its minimum glibc version to 2.17](https://github.com/rust-lang/rust/pull/95026). Unless it's a major obstacle, I'd prefer to have cap-std continue to support the versions of glibc supported by Rust.

cap-tempfile/src/tempfile.rs
```diff
@@ -0,0 +1,251 @@
...
+                generate_name_in(self.dir, &self.fd)?
+            } else {
+                // We can only have an anonymous file at this point on Linux.
+                unreachable!()
+            }
+        };
+        // And try the rename into place.
+        rustix::fs::renameat(self.dir, &tempname, self.dir, destname).map_err(|e| {
+            // But, if we catch an error here, then move ownership back into self,
+            // whioch means the Drop invocation will clean it up.
```
sunfishcode, line 133: Typo: whioch

cap-tempfile/src/tempfile.rs
```diff
@@ -0,0 +1,251 @@
...
+/// Assign a random name to a currently anonymous O_TMPFILE descriptor.
+#[cfg(any(target_os = "android", target_os = "linux"))]
+fn generate_name_in(subdir: &Dir, f: &File) -> io::Result<String> {
+    use rustix::fd::AsFd;
+    use rustix::fs::AtFlags;
+    let procself_fd = rustix::io::proc_self_fd()?;
+    let fdnum = rustix::path::DecInt::from_fd(&f.as_fd());
+    let fdnum = fdnum.as_c_str();
+    super::retry_with_name_ignoring(io::ErrorKind::AlreadyExists, |name| {
+        rustix::fs::linkat(&procself_fd, fdnum, subdir, name, AtFlags::SYMLINK_FOLLOW)
```
sunfishcode, line 60: Is `SYMLINK_FOLLOW` the right thing to do here? The name is randomly generated in order to try to avoid conflicting with any pre-existing file; it seems like if there's a pre-existing symlink, you'd want to just fail and try a different name too.

cap-tempfile/src/tempfile.rs
```diff
@@ -0,0 +1,251 @@
...
+    if cfg!(any(target_os = "android", target_os = "linux")) {
+        if let Some(f) = new_tempfile_linux(d)? {
+            return Ok((f, None));
+        }
+    }
+    // Otherwise, fall back to just creating a randomly named file.
+    let mut opts = cap_std::fs::OpenOptions::new();
+    opts.read(true);
+    opts.write(true);
+    opts.create_new(true);
```
sunfishcode, line 78: As above, it seems like this should also do `.custom_flags(O_NOFOLLOW)` or so.

cap-tempfile/src/tempfile.rs
cgwalters: Unfortunately...we need this flag in order for the *source* path of `/proc/self/fd/N` to be followed.  Though, `AT_EMPTY_PATH` claims it supports this case, I'm not getting it to work in some quick testing.  Will look at this more closely.

sunfishcode: `AT_EMPTY_PATH` requires `CAP_DAC_READ_SEARCH`, which is common to not have.

cgwalters: I double checked that `linkat()` will not follow target symlinks.

cgwalters:
OK, MacOS and Windows should compile at least now.

cgwalters:
> According to the linked bug report, the bug is present in glibc 2.20, however Rust itself is only now https://github.com/rust-lang/rust/pull/95026. Unless it's a major obstacle, I'd prefer to have cap-std continue to support the versions of glibc supported by Rust.

Yeah, fair.  Though it should be noted, it looks like the [tempfile crate uses this unconditionally today](https://github.com/Stebalien/tempfile/blob/1a40687e06eb656044e3d2dffa1379f04b3ef3fd/src/file/imp/unix.rs#L77).

(We need to match the same errno fallbacks too, will look at that)

Perhaps one possible trick here could be to use a weak symbol to detect if we have a new enough glibc?  I'd guess rustix could expose some API for this, given that it's effectively doing it internally; for example if we have `getrandom` we know we have a new enough glibc.

cgwalters:
> Perhaps one possible trick here could be to use a weak symbol to detect if we have a new enough glibc?

Or we could use `openat2` - we already have all the infrastructure for that.  The downside though is this would exclude systems that have new enough Linux for `O_TMPFILE` but not `openat2`.  But OTOH people using `cap-std` on such old systems are already going to be paying a high price and this would only be another small hit.

cgwalters:
> Or we could use openat2 

I think to do that nicely we'd want to make `open_beneath` from `cap-primitives` be `pub` or so?  There are important things like handling android and seccomp there.

sunfishcode:
(Don't worry about the windows-2016 failure btw.)

I think this PR is in good shape; I just want to figure out the `O_TMPFILE` situation here.

> > According to the linked bug report, the bug is present in glibc 2.20, however Rust itself is only now [rust-lang/rust#95026](https://github.com/rust-lang/rust/pull/95026). Unless it's a major obstacle, I'd prefer to have cap-std continue to support the versions of glibc supported by Rust.
> 
> Yeah, fair. Though it should be noted, it looks like the [tempfile crate uses this unconditionally today](https://github.com/Stebalien/tempfile/blob/1a40687e06eb656044e3d2dffa1379f04b3ef3fd/src/file/imp/unix.rs#L77).

If I understand correctly, that code in the tempfile is using `open` rather than `openat`, so on x86_64 and possibly others the bug is hidden in practice (according to the glibc bug linked above).

> Perhaps one possible trick here could be to use a weak symbol to detect if we have a new enough glibc? I'd guess rustix could expose some API for this, given that it's effectively doing it internally; for example if we have `getrandom` we know we have a new enough glibc.

My sense here is that this is something that would be better to handle internally in rustix, rather than exposing an API. Detecting it by using the `weak!` macro looking for `getrandom` sounds reasonable to me, and for a fallback when we don't have a new enough glibc, rustix could use `libc::syscall(SYS_openat, ...) `, which avoids the bug. Does that sound workable?

cgwalters:
> If I understand correctly, that code in the tempfile is using open rather than openat, so on x86_64 and possibly others the bug is hidden in practice (according to the glibc bug linked above).

Ah yes, I think you're right.

> My sense here is that this is something that would be better to handle internally in rustix, rather than exposing an API. Detecting it by using the weak! macro looking for getrandom sounds reasonable to me, and for a fallback when we don't have a new enough glibc, rustix could use libc::syscall(SYS_openat, ...) , which avoids the bug. Does that sound workable?

Sounds good to me!

cgwalters:
>  Detecting it by using the weak! macro looking for getrandom sounds reasonable to me

I briefly looked at this, and today the `weak!` is wrapped up in `weak_or_syscall!`, and I couldn't figure out an elegant way to make the `weak!` inside there conditionally `pub(crate)` or so.

Here's my hack to duplicate it:

```
From 84dfe8d0600a39c907a887ae7237a5f24a6410d5 Mon Sep 17 00:00:00 2001
From: Colin Walters <walters@verbum.org>
Date: Fri, 1 Apr 2022 08:45:09 -0400
Subject: [PATCH] wip

---
 src/imp/libc/mod.rs           | 6 ++++++
 src/imp/libc/rand/syscalls.rs | 3 +++
 src/imp/libc/weak.rs          | 4 ++--
 3 files changed, 11 insertions(+), 2 deletions(-)

diff --git a/src/imp/libc/mod.rs b/src/imp/libc/mod.rs
index 96a040b..1fbcbbd 100644
--- a/src/imp/libc/mod.rs
+++ b/src/imp/libc/mod.rs
@@ -62,3 +62,9 @@ pub(crate) mod rand;
 pub(crate) mod thread;
 #[cfg(not(windows))]
 pub(crate) mod time;
+
+/// If the host libc is glibc, detect if it's at least version 2.25.
+#[cfg(not(windows))]
+pub(crate) fn is_atleast_glibc_2_25() -> bool {
+    rand::syscalls::libc_getrandom_weak.get().is_some()
+}
diff --git a/src/imp/libc/rand/syscalls.rs b/src/imp/libc/rand/syscalls.rs
index e860f40..380f643 100644
--- a/src/imp/libc/rand/syscalls.rs
+++ b/src/imp/libc/rand/syscalls.rs
@@ -5,6 +5,9 @@
 #[cfg(target_os = "linux")]
 use {super::super::c, super::super::conv::ret_ssize_t, crate::io, crate::rand::GetRandomFlags};
 
+// A duplicate weak symbol because we use this to detect the glibc version.
+weak! { pub(crate) fn libc_getrandom_weak(*mut c::c_void, c::size_t, c::c_uint) -> c::ssize_t }
+
 #[cfg(target_os = "linux")]
 pub(crate) fn getrandom(buf: &mut [u8], flags: GetRandomFlags) -> io::Result<usize> {
     // `getrandom` wasn't supported in glibc until 2.25.
diff --git a/src/imp/libc/weak.rs b/src/imp/libc/weak.rs
index aecc081..203969f 100644
--- a/src/imp/libc/weak.rs
+++ b/src/imp/libc/weak.rs
@@ -36,9 +36,9 @@ const NULL: *mut c_void = null_mut();
 const INVALID: *mut c_void = 1 as *mut c_void;
 
 macro_rules! weak {
-    (fn $name:ident($($t:ty),*) -> $ret:ty) => (
+    ($vis:vis fn $name:ident($($t:ty),*) -> $ret:ty) => (
         #[allow(non_upper_case_globals)]
-        static $name: $crate::imp::weak::Weak<unsafe extern fn($($t),*) -> $ret> =
+        $vis static $name: $crate::imp::weak::Weak<unsafe extern fn($($t),*) -> $ret> =
             $crate::imp::weak::Weak::new(concat!(stringify!($name), '\0'));
     )
 }
-- 
2.35.1

```

cgwalters:
OK moving that bit to https://github.com/bytecodealliance/rustix/pull/268

sunfishcode:
https://github.com/bytecodealliance/rustix/pull/268 is now released, in rustix 0.33.6.

cgwalters:
Anything else that needs doing from my side for this?

sunfishcode:
Ah, thanks for the ping. Looks like this is ready to go, the CI failure is just windows-2016 which is fixed on main.

PR #240, opened 2022-03-31
tempfile: Move tests into a `#[cfg(test)] mod test`

cgwalters:
I think having unit tests inside a module is the standard; I noticed
that this project doesn't seem to do that consistently - is there
a reason for that?

As I understand it, the main value of `#[cfg(test)] mod test {}` is
that it avoids every user of the crate pointlessly compiling the tests.

### DIFF
--- a/cap-tempfile/src/lib.rs
+++ b/cap-tempfile/src/lib.rs
@@ -177,61 +177,66 @@ pub fn tempdir_in(dir: &Dir) -> io::Result<TempDir> {
     TempDir::new_in(dir)
 }
 
-#[test]
-fn drop_tempdir() {
-    use crate::ambient_authority;
+#[cfg(test)]
+mod test {
+    use super::*;
 
-    let t = tempdir(ambient_authority()).unwrap();
-    drop(t)
-}
+    #[test]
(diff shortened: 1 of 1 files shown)


sunfishcode:
It's my understanding that `#[test]` functions aren't compiled by default. Adding this:
```rust
#[test]
fn it_works() {
    does_not_exist();
}
```
seems to confirm this.

With this understanding, my habit is to skip `mod test {}` because it adds clutter and indentation without adding value in most cases. But, I'd be happy to learn about other considerations.

cgwalters:
So `#[test]` seems to also implicitly be `#[cfg(test)]` - which makes a ton of sense.  But then why is this pattern in the book? https://doc.rust-lang.org/book/ch11-01-writing-tests.html ?

I'm curious now and will see if I can find out.

cgwalters:
xref https://rust-lang.zulipchat.com/#narrow/stream/122651-general/topic/why.20mod.20test.3F

sunfishcode:
Also of potential interest here: https://github.com/rust-lang/cargo/issues/10559

PR #244, opened 2022-05-01
tempfile: Add missing `Debug` derive

cgwalters:
Hit this when trying to embed in another struct.

### DIFF
--- a/cap-tempfile/src/tempfile.rs
+++ b/cap-tempfile/src/tempfile.rs
@@ -2,6 +2,7 @@
 
 use cap_std::fs::{Dir, File};
 use std::ffi::OsStr;
+use std::fmt::Debug;
 use std::io::{self, Write};
 use std::io::{Read, Seek};
 
@@ -38,6 +39,14 @@ pub struct TempFile<'d> {
     dir: &'d Dir,
     fd: File,
     name: Option<String>,
+}
+
+impl<'d> Debug for TempFile<'d> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        // Manual Debug implementation to omit the file reference and name so
+        // we don't leak the path, the same as `cap_std::fs::File`.
+        f.debug_struct("TempFile").field("dir", &self.dir).finish()
+    }
 }
 
 #[cfg(any(target_os = "android", target_os = "linux"))]


sunfishcode:
cap-std currently has the property that `Debug` impls shouldn't expose path strings that are otherwise encapsulated. For example, [`cap_std::fs::File`'s `Debug` impl](https://github.com/bytecodealliance/cap-std/blob/main/cap-std/src/fs/file.rs#L570) doesn't print the file's path, even though `std::fs::File`'s does. This is something of a theoretical concern, and I'm open to discussing whether this policy is really necessary :-). But otherwise, would you mind changing this to a manual `Debug` impl that excludes `self.name`?

cgwalters:
Hmm.  I see the idea, but OTOH anyone who *really* wants a path can pretty easily construct `/proc/self/fd/N` on their own too.

But anyways updated to do so!

cgwalters:
Did you consider changing `std` to omit the path for `File`?  Some quick digging with git leads to this having originated in https://github.com/rust-lang/rust/issues/24570#issuecomment-94215898 which seems kind of like a whim, like "why not do this"?  But capability orientation seems like a reason why not to do it.

And that commenter is still around in the Rust community.

sunfishcode:
> Hmm. I see the idea, but OTOH anyone who really wants a path can pretty easily construct /proc/self/fd/N on their own too.

/proc/self/fd/N is an absolute path, so [this clippy config](https://github.com/sunfishcode/ambient-authority/blob/main/clippy.toml) can detect code that does that.

On the other hand, the clippy lint isn't completely watertight, so it's debatable whether these `Debug` impls should really be this strict. On the first hand though, I could imagine someday building more powerful tools. On the other hand again, those don't exist yet.

> Did you consider changing std to omit the path for File? Some quick digging with git leads to this having originated in https://github.com/rust-lang/rust/issues/24570#issuecomment-94215898 which seems kind of like a whim, like "why not do this"? But capability orientation seems like a reason why not to do it.

Printing the path does turn out to be a useful feature. When you're looking at debug log output, a filename can be more meaningful than just "fd 27" or whatever. And in a regular `std::fs`-using codebase, paths aren't very special, so why not include them?

In a `cap_std::fs`-using codebase, perhaps users are less oriented around paths, and perhaps omitting the path will help encourage them to be even more so. Admittedly, this is also theoretical reasoning.

sunfishcode:
This is now released in 0.24.4.

PR #246, opened 2022-05-02
tempfile: Some doc tweaks

cgwalters:
I noticed that the `tempfile::tempfile` link was broken when
looking online; fixed that and did some other tweaks while I had
the file open.


sunfishcode:
:+1:

PR #247, opened 2022-05-26
Add opt-in tokio 1.0 support to cap-async-std

definitelynobody:
Just a small change to allow cap-async-std to run under a tokio 1.0 runtime.


sunfishcode:
Thanks!

sunfishcode:
This is now released in 0.24.4.

PR #248, opened 2022-06-03
Return file creation times on Linux (using statx)

ongardie:
Rust's standard library returns file creation times on Linux (at least with glibc), and now cap-std will too.

File creation times are available on Linux via the `statx` syscall (not `fstatat`), introduced in kernel 4.11, which was released in April 2017. The Rust standard library already uses `statx` on Linux with glibc.

Making cap-std support file creation times on Linux required two changes:

1. `File::metadata()` uses the standard library (which uses `statx`), but this wouldn't ever copy over the created field on Linux. Now, any time the created field is set in the std Metadata struct, it's also set in the cap-primitives Metadata, regardless of platform.

2. `stat_unchecked` is used in several places, including fetching DirEntry metadata. Before, it called `fstatat` directly. Now, it calls `statx` on Linux when available, and it falls back to `fstatat` otherwise. Fortunately, Dan Gohman (@sunfishcode) had already added a method to convert `statx` results in commit d1fa735 (PR #105) in 2020.

This commit also adds a new test to make sure file creation times are set in cap-std Metadata if they are set in std Metadata.

### DIFF
--- a/cap-primitives/src/rustix/fs/stat_unchecked.rs
+++ b/cap-primitives/src/rustix/fs/stat_unchecked.rs
@@ -23,9 +23,9 @@ pub(crate) fn stat_unchecked(
         FollowSymlinks::No => AtFlags::SYMLINK_NOFOLLOW,
     };
 
-    // `statx` is preferred on Linux because it can return creation times. Linux
-    // kernels prior to 4.11 don't have `statx` and return `ENOSYS`. We store the
-    // availability in a global to avoid unnecessary syscalls.
+    // `statx` is preferred on Linux because it can return creation times.
+    // Linux kernels prior to 4.11 don't have `statx` and return `ENOSYS`. We
+    // store the availability in a global to avoid unnecessary syscalls.
     #[cfg(all(target_os = "linux", target_env = "gnu"))]
     static HAS_STATX: AtomicBool = AtomicBool::new(true);
 
--- a/cap-primitives/src/rustix/fs/stat_unchecked.rs
+++ b/cap-primitives/src/rustix/fs/stat_unchecked.rs
@@ -9,7 +9,7 @@ use std::{fs, io};
 #[cfg(all(target_os = "linux", target_env = "gnu"))]
 use rustix::fs::{statx, StatxFlags};
 #[cfg(all(target_os = "linux", target_env = "gnu"))]
-use std::sync::atomic::{AtomicBool, Ordering};
+use std::sync::atomic::{AtomicU8, Ordering};
 
 /// *Unsandboxed* function similar to `stat`, but which does not perform
 /// sandboxing.
@@ -24,23 +24,51 @@ pub(crate) fn stat_unchecked(
     };
 
     // `statx` is preferred on Linux because it can return creation times.
-    // Linux kernels prior to 4.11 don't have `statx` and return `ENOSYS`. We
-    // store the availability in a global to avoid unnecessary syscalls.
+    // Linux kernels prior to 4.11 don't have `statx` and return `ENOSYS`.
+    // Older versions of Docker/seccomp would return `EPERM` for `statx`; see
+    // <https://github.com/rust-lang/rust/pull/65685/>. We store the
+    // availability in a global to avoid unnecessary syscalls.
     #[cfg(all(target_os = "linux", target_env = "gnu"))]
-    static HAS_STATX: AtomicBool = AtomicBool::new(true);
-
-    #[cfg(all(target_os = "linux", target_env = "gnu"))]
-    if HAS_STATX.load(Ordering::Relaxed) {
-        let statx_result = statx(
-            start,
-            path,
-            atflags,
-            StatxFlags::BASIC_STATS | StatxFlags::BTIME,
-        );
-        match statx_result {
-            Ok(statx) => return Ok(MetadataExt::from_rustix_statx(statx)),
-            Err(rustix::io::Error::NOSYS) => HAS_STATX.store(false, Ordering::Relaxed),
-            Err(e) => return Err(e.into()),
+    {
+        // 0: Unknown
+        // 1: Not available
+        // 2: Available
+        static STATX_STATE: AtomicU8 = AtomicU8::new(0);
+        let state = STATX_STATE.load(Ordering::Relaxed);
+        if state != 1 {
+            let statx_result = statx(
+                start,
+                path,
+                atflags,
+                StatxFlags::BASIC_STATS | StatxFlags::BTIME,
+            );
+            match statx_result {
+                Ok(statx) => {
+                    if state == 0 {
+                        STATX_STATE.store(2, Ordering::Relaxed);
+                    }
+                    return Ok(MetadataExt::from_rustix_statx(statx));
+                }
+                Err(rustix::io::Error::NOSYS) => STATX_STATE.store(1, Ordering::Relaxed),
+                Err(rustix::io::Error::PERM) if state == 0 => {
+                    // This is an unlikely case, as `statx` doesn't normally
+                    // return `PERM` errors. One way this can happen is when
+                    // running on old versions of seccomp/Docker. If `statx` on
+                    // the current working directory returns a similar error,
+                    // then stop using `statx`.
+                    if let Err(rustix::io::Error::PERM) = statx(
+                        rustix::fs::cwd(),
+                        "",
+                        AtFlags::EMPTY_PATH,
+                        StatxFlags::empty(),
+                    ) {
+                        STATX_STATE.store(1, Ordering::Relaxed);
+                    } else {
+                        return Err(rustix::io::Error::PERM.into());
+                    }
+                }
+                Err(e) => return Err(e.into()),
+            }
         }
     }
 
--- a/cap-primitives/src/rustix/fs/stat_unchecked.rs
+++ b/cap-primitives/src/rustix/fs/stat_unchecked.rs
@@ -37,10 +37,10 @@ pub(crate) fn stat_unchecked(
             atflags,
             StatxFlags::BASIC_STATS | StatxFlags::BTIME,
         );
-        if let Err(rustix::io::Error::NOSYS) = statx_result {
-            HAS_STATX.store(false, Ordering::Relaxed);
-        } else {
-            return Ok(statx_result.map(MetadataExt::from_rustix_statx)?);
+        match statx_result {
+            Ok(statx) => return Ok(MetadataExt::from_rustix_statx(statx)),
+            Err(rustix::io::Error::NOSYS) => HAS_STATX.store(false, Ordering::Relaxed),
+            Err(e) => return Err(e.into()),
         }
     }
 
--- a/cap-primitives/src/rustix/fs/stat_unchecked.rs
+++ b/cap-primitives/src/rustix/fs/stat_unchecked.rs
@@ -3,6 +3,14 @@ use rustix::fs::{statat, AtFlags};
 use std::path::Path;
 use std::{fs, io};
 
+// TODO: update all these to
+// #[cfg(any(target_os = "android", target_os = "linux"))]
+// once we're on restix >= v0.34.3.
+#[cfg(all(target_os = "linux", target_env = "gnu"))]
+use rustix::fs::{statx, StatxFlags};
+#[cfg(all(target_os = "linux", target_env = "gnu"))]
+use std::sync::atomic::{AtomicBool, Ordering};
+
 /// *Unsandboxed* function similar to `stat`, but which does not perform
 /// sandboxing.
 pub(crate) fn stat_unchecked(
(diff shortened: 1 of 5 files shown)


cap-primitives/src/rustix/fs/stat_unchecked.rs
```diff
@@ -15,5 +23,26 @@ pub(crate) fn stat_unchecked(
         FollowSymlinks::No => AtFlags::SYMLINK_NOFOLLOW,
     };
 
+    // stax is preferred on Linux because it can return creation times. Linux
```
sunfishcode, line 26:
```suggestion
    // `statx` is preferred on Linux because it can return creation times. Linux
```

cap-primitives/src/rustix/fs/stat_unchecked.rs
```diff
@@ -15,5 +23,26 @@ pub(crate) fn stat_unchecked(
...
+    // kernels prior to 4.11 don't have statx and return ENOSYS. We store the
```
sunfishcode, line 27:
```suggestion
    // kernels prior to 4.11 don't have `statx` and return `ENOSYS`. We store the
```

cap-primitives/src/rustix/fs/stat_unchecked.rs
```diff
@@ -15,5 +23,26 @@ pub(crate) fn stat_unchecked(
...
+            start,
+            path,
+            atflags,
+            StatxFlags::BASIC_STATS | StatxFlags::BTIME,
+        );
+        if let Err(rustix::io::Error::NOSYS) = statx_result {
+            HAS_STATX.store(false, Ordering::Relaxed);
+        } else {
+            return Ok(statx_result.map(MetadataExt::from_rustix_statx)?);
+        }
```
sunfishcode, line 44: As a minor style nit, I prefer to do a single `match` on an `rustix::io::Result`, rather than picking out one error case with an `if let` and handling other error cases with `.map`, like this:
```rust
match statx_result {
  Ok(statx) => return Ok(...),
  Err(rustix::io::Error::NOSYS) => ...
  Err(e) => return Err(e),
}
```

tests/metadata-ext.rs
```diff
@@ -83,3 +83,55 @@ fn test_metadata_ext() {
...
+    let tmpdir = tmpdir();
+    let a = check!(tmpdir.create("a"));
+    let a_metadata = check!(a.metadata());
+
+    let modified = check!(a_metadata.modified());
+    let tolerance = std::time::Duration::from_secs(10);
+    let expected = (modified - tolerance)..(modified + tolerance);
+
+    // If the standard library supports file creation times, then cap-std
+    // should too.
```
sunfishcode, line 98: Nice, I like this correspondence.

tests/metadata-ext.rs
```diff
@@ -83,3 +83,55 @@ fn test_metadata_ext() {
...
+    let std_supports_created = matches!(
+        a.into_std().metadata(),
+        Ok(m) if m.created().is_ok(),
+    );
```
sunfishcode, line 102: Instead of using the `modified()` timestamp for `expected`, could you use the `created()` timestamp for `expected`? And for the directories, could we always call `std::fs::metadata` so that we always get the `created()` timestamp from std? It seems like that should eliminate the need for a tolerance, because the creation time from cap-std should be exactly the same as the creation time from std.

sunfishcode (commented):
Looks good, thanks for working on this!

cap-primitives/src/rustix/fs/stat_unchecked.rs
ongardie: I went back and forth on that, but the single match is probably better. Updating.

tests/metadata-ext.rs
```diff
@@ -83,3 +83,55 @@ fn test_metadata_ext() {
         );
     }
 }
+
+#[test]
+fn test_metadata_ext_created() {
```
ongardie, line 88: Is this test in the right / a reasonable place? I wasn't sure.

cap-primitives/src/rustix/fs/stat_unchecked.rs
```diff
@@ -15,5 +23,26 @@ pub(crate) fn stat_unchecked(
...
+    if HAS_STATX.load(Ordering::Relaxed) {
+        let statx_result = statx(
+            start,
+            path,
+            atflags,
+            StatxFlags::BASIC_STATS | StatxFlags::BTIME,
+        );
+        match statx_result {
+            Ok(statx) => return Ok(MetadataExt::from_rustix_statx(statx)),
+            Err(rustix::io::Error::NOSYS) => HAS_STATX.store(false, Ordering::Relaxed),
```
ongardie, line 42: I saw a comment in `open_impl.rs` about seccomp on Android making ENOSYS a bad idea. I don't think that applies here yet, but it might later. Is there a better approach? Should I place a comment on this landmine?

cap-primitives/src/rustix/fs/stat_unchecked.rs
```diff
@@ -3,6 +3,14 @@ use rustix::fs::{statat, AtFlags};
 use std::path::Path;
 use std::{fs, io};
 
+// TODO: update all these to
+// #[cfg(any(target_os = "android", target_os = "linux"))]
+// once we're on restix >= v0.34.3.
```
ongardie, line 8: Any idea when the rustix version will be updated? I don't have a sense of how big that is, so I didn't want to add it to this PR.

tests/metadata-ext.rs
sunfishcode: Ah, good question. metadata-ext.rs is for `MetadataExt` APIs, but the `created()` field is a regular `Metadata` API. This can go in `fs_additional.rs`.

cap-primitives/src/rustix/fs/stat_unchecked.rs
ongardie: Updated in c7f65a2.

cap-primitives/src/rustix/fs/stat_unchecked.rs
sunfishcode: The biggest thing for cap-std is that there are some semver incompatibilities between versions of io-lifetimes, which is tracking changes from [I/O safety stabilization](https://github.com/rust-lang/rust/issues/87074). Fortunately, the FCP looks to be almost done. Once it's done, I plan to do an io-lifetimes release and then a [rustix 0.35 release](https://github.com/bytecodealliance/rustix/issues/313), and then update cap-std and various other things.

cap-primitives/src/rustix/fs/stat_unchecked.rs
sunfishcode: Rust's own standard library calls `statx` without doing an explicit version check, so we're ok on Androidd here.

Looking into it though, I notice that there is a potential problem with `statx` returning `EPERM` on old versions of Docker; std does a [special check](https://github.com/rust-lang/rust/pull/65685/) to avoid this problem. If you want to implement that same trick here, that'd be great, though since rustix [also needs to do this](https://github.com/bytecodealliance/rustix/issues/352), it's also fine to leave this code as-is for now, and I'll handle this when I update rustix.

tests/metadata-ext.rs
ongardie: Yep, that makes more sense than what I had done. Updated.

tests/metadata-ext.rs
ongardie: Updated.

tests/fs_additional.rs
```diff
@@ -872,3 +872,37 @@ fn reopen_fd() {
...
+        let mut entries = check!(dir.entries());
+        let entry = check!(entries.next().unwrap());
+        assert_eq!(entry.file_name(), "file");
+        assert!(entries.next().is_none(), "unexpected dir entry");
+        check!(entry.metadata())
+    };
+
+    let std_dir = check!(dir.into_std_file().metadata());
+    let std_file = check!(file.into_std().metadata());
+
```
ongardie, line 895: This test is much nicer now, I think. However, it's a little funny as is. It does all this setup to extract metadata in various ways from cap_std and std, but then it only compares the created times of the metadata and nothing else. Thoughts?

cap-primitives/src/rustix/fs/stat_unchecked.rs
ongardie: Nice catch. It seems like if we merge this PR now, it'd break metadata for not-very old versions of Docker, so that's probably bad. 

Why/where does rustix need this logic? I thought it just wrapped `statx` and doesn't really care whether or not `statx` succeeds.

tests/fs_additional.rs
sunfishcode: On one hand, creation times have an interesting enough implementation landscape that I feel it's justified to have a fair amount of test setup just for that one purpose, so this seems fine. On the other, since we have all this setup here, it'd be nice to expand this to test all of `Metadata`'s accessors to make sure they all match. `accessed()` might need to be a `>=` check instead of equality because files can get spurious access-time updates.

I don't want to bog you down if you're working on a project which just needs the creation times fix. So I'll leave it up to you for now.

cap-primitives/src/rustix/fs/stat_unchecked.rs
sunfishcode: Rustix uses `statx` in its implementation of `rustix::fs::stat`. Rustix usually doesn't do that kind of emulation, but in this case, it's motivated by a desire to be y2038-ready on platforms which support that. On Linux, that means using `statx` instead of the native `stat` family of syscalls whenever possible. So if you want to fix this now, that'd be great, but I am also happy to fix this myself at the same time that I fix rustix.

ongardie: Ah, I didn't see that earlier because I was searching through an older version, oops.

tests/fs_additional.rs
```diff
@@ -872,3 +872,114 @@ fn reopen_fd() {
...
+    }
+
+    assert_eq!(std.len(), cap.len());
+
+    assert_eq!(std.permissions().readonly(), cap.permissions().readonly());
+    if cfg!(unix) {
+        use std::os::unix::fs::PermissionsExt;
+        // The standard library returns file format bits with `mode()`, whereas
+        // cap-std currently doesn't.
+        assert_eq!(std.permissions().mode() & 0o7777, cap.permissions().mode());
```
ongardie, line 934: This was unexpected and a bit surprising to me. Is it intentional? If so, is that documented somewhere?

tests/fs_additional.rs
ongardie: Ok, I expanded this out for all of the accessible Metadata (with only Unix extensions for now).

tests/fs_additional.rs
```diff
@@ -872,3 +872,114 @@ fn reopen_fd() {
...
+    }
+
+    // If the standard library supports file modified/accessed/created times,
+    // then cap-std should too.
+    if let Ok(expected) = std.modified() {
+        assert_eq!(expected, check!(cap.modified()).into_std());
+    }
+    // The access times might be a little different due to either our own
+    // or concurrent accesses.
+    const ACCESS_TOLERANCE_SEC: u32 = 60;
```
ongardie, line 944: I still don't love this but didn't want to assume we made the cap or std access first, and I wanted to check the access times weren't wildly far into the future either. I guess it's pretty easy to find and update later if it causes trouble.

cap-primitives/src/rustix/fs/stat_unchecked.rs
ongardie: I just simulated the seccomp statx -> EPERM behavior on my system using this Python script (as root):
```python
import errno
import os
import seccomp

f = seccomp.SyscallFilter(seccomp.ALLOW)
f.add_rule(seccomp.ERRNO(errno.EPERM), "statx")
f.load()

os.execl(
    "target/debug/deps/fs_additional-d7e3e72f4c824747",
    "fs_additional",
    "--nocapture",
    "--test-threads=1",
    "metadata_vs_std_fs",
)
```

It's not quite as accurate as getting an old Docker version, but I'd rather not do that :)

ongardie: I updated `stat_unchecked` to handle `EPERM`. I couldn't do the same `EFAULT` trick as Rust `std` uses (I think it'd require unsafe), but I pushed the test off the critical path. Let me know what you think.

tests/fs_additional.rs
sunfishcode: It's cap-std's intention to match std's behavior at least, so it looks like a bug in cap-std. I filed https://github.com/bytecodealliance/cap-std/issues/249 to track this.

cap-primitives/src/rustix/fs/stat_unchecked.rs
sunfishcode: I'm considering adding something in Rustix to allow or the null-pointer `EFAULT` trick, but for now the code you have here looks good.

ongardie:
Thanks for all your guidance with this, @sunfishcode, especially since there's a lot of relevant context that I wasn't aware of when I started. I think the code improved with each iteration, and it's nice that the expanded test caught another issue right away. Cheers.

sunfishcode:
This is now released in cap-std 0.25.

PR #252, opened 2022-06-16
cap-primitives: Add `read_link_contents` and `symlink_contents`

cgwalters:
There are use cases for reading and writing the contents of
symbolic links without the check that errors out on absolute paths.
It's safe when the calling process is not going to interpret it.

For example in rpm-ostree, we construct a new root filesystem which
in some cases may have absolute links, but the calling process doesn't
read those links back.  Instead we spawn containers into the target
root to run code there.

---

### DIFF
--- a/cap-primitives/Cargo.toml
+++ b/cap-primitives/Cargo.toml
@@ -21,6 +21,9 @@ fs-set-times = "0.17.0"
 io-extras = "0.15.0"
 io-lifetimes = { version = "0.7.0", default-features = false }
 
+[dev-dependencies]
+cap-tempfile = { path = "../cap-tempfile" }
+
 [target.'cfg(not(windows))'.dependencies]
 rustix = { version = "0.35.6", features = ["fs", "process", "procfs", "termios", "time"] }
 
--- a/cap-primitives/src/fs/mod.rs
+++ b/cap-primitives/src/fs/mod.rs
@@ -80,7 +80,7 @@ pub use open_dir::*;
 pub use open_options::OpenOptions;
 pub use permissions::Permissions;
 pub use read_dir::{read_base_dir, read_dir, ReadDir};
-pub use read_link::read_link;
+pub use read_link::{read_link, read_link_contents};
 pub use remove_dir::remove_dir;
 pub use remove_dir_all::remove_dir_all;
 pub use remove_file::remove_file;
@@ -92,7 +92,7 @@ pub use set_permissions::set_permissions;
 pub use set_times::{set_times, set_times_nofollow};
 pub use stat::stat;
 #[cfg(not(windows))]
-pub use symlink::symlink;
+pub use symlink::{symlink, symlink_contents};
 #[cfg(windows)]
 pub use symlink::{symlink_dir, symlink_file};
 pub use system_time_spec::SystemTimeSpec;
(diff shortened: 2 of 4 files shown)


cgwalters:
Ah yes, Windows.  Will look...

sunfishcode:
I took the liberty of fixing the Windows error here; it just needed an `Into::into` in the right place.

That said, while doing this, I noticed that [`std::fs::read_link`](https://doc.rust-lang.org/stable/std/fs/fn.read_link.html) returns a `PathBuf`, which leads me to wonder whether we really need to use `OsString` here, given that Rust itself doesn't. You're right that symbolic link payloads can be almost arbitrary data, however Unix paths can be almost arbitrary data too, and `PathBuf` can represent them, so it seems fine here too. Are there other considerations I'm unaware of here?

cgwalters:
> That said, while doing this, I noticed that [std::fs::read_link](https://doc.rust-lang.org/stable/std/fs/fn.read_link.html) returns a PathBuf, which leads me to wonder whether we really need to use OsString here, given that Rust itself doesn't.

OK, totally fair point.  I've:

- dropped those prep commits
- changed the `read_link_contents` API to return `PathBuf`
- Also added a `symlink_contents` API both on general principle, and so we can unit test `read_link_contents`

cgwalters:
(BTW, are we agreed to add new `read_link_contents` and `symlink_contents` to `cap_std::fs::Dir` too?)

> Good question. I would prefer to maintain the property that functions that are supported in WASI and non-WASI behave the same. That is, it's fine if we have some extra functions that don't work in WASI, like read_link_contents, but I would prefer to avoid having read_link work one way on WASI and another on native.

The way I'd say this is - we are imposing a somewhat high cost on Unix applications porting from Rust std to cap-std in that the `read_link` method doesn't quite correspond in a less-than-obvious way.

What I honestly don't know is: how important is this check for WASI?  Is it just...something you added proactively on a whim?  Did it come out of an actual vulnerability?  Somewhere in between?

Having a new API for this seems OK to me in the end; most codebases that will port to std :arrow_right: cap-std will have a fair amount of nontrivial adjustment anyways.  But the cases where the method is named the same, but silently does something different have a high cost.

cgwalters:
> I think this is related to the RESOLVE_BENEATH vs. RESOLVE_IN_ROOT difference we discussed earlier. As I understand it, in your use case, a Dir can represent directory with a root-filesystem layout, where absolute paths have a clear meaning, and you accept the constraint that everyone following the symlinks should be careful to use RESOLVE_IN_ROOT or chroot behavior.

Yes.   (Though I'd s/chroot/containerization/ - no one should be using chroot nowadays, every operating system worth using has a better API)

> In WASI, and in the other cap-std use cases I'm aware of, absolute paths don't have a clear meaning, and do have potential risks. 

Beyond just the "constructing and introspecting an alternative root filesystem" case, I think a good example of a cap-std use case is e.g. the [tar crate](https://docs.rs/tar/latest/tar/) wanting to serialize a filesystem into a tarball - as well as deserialize.  We don't want to ever accidentally traverse outside of the target directory, but we do want to faithfully read and write absolute links.

> These two perspectives have a lot of overlap, so my guess is that we can by addressing it in things like read_link_contents and symlink_contents.

Yeah, I am fine with this too - to confirm, once the cap-primitives API is released, are you OK to add this (in all the variants, utf8 and async etc.) to cap-std?

PR #256, opened 2022-06-21
Don't apply a mask on the flags in `PermissionsExt`.

### LINKED ISSUE #249: PermissionsExt::mode masks off the format bits
As reported [here](https://github.com/bytecodealliance/cap-std/pull/248/files#r889355402), cap-std's `PermissionExt::mode` returns a mode value with the format bits masked out, while std's `PermissionsExt::mode` returns the raw mode value. It's cap-std'd intention to match std, so it should remove the masking.

sunfishcode:
For consistency with std, don't apply a mask to the flags bits in
`PermissionsExt`.

Fixes #249.


sunfishcode:
I find this a little surprising, but it's what std does, and cap-std intends to match std.

PR #265, opened 2022-06-27
tempfile: Re-export `cap-std`

cgwalters:
In many of the projects I work on we use Github dependabot, configured
to open a separate PR for each bump.  But right now dependabot
does not understand the cargo/rust level dependencies involved in e.g.
https://github.com/coreos/cap-std-ext/blob/main/Cargo.toml#L12-L14

I think it's general best practice to re-export dependencies used
in public API, so that consumers can keep their direct set
smaller, and this way dependabot updates in this form will just work
because we can just depend on `cap-tempfile`.

### DIFF
--- a/cap-tempfile/src/lib.rs
+++ b/cap-tempfile/src/lib.rs
@@ -21,6 +21,9 @@ pub mod utf8;
 mod tempfile;
 pub use tempfile::*;
 
+/// Re-export because we use this in our public API.
+pub use cap_std;
+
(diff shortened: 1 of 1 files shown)


cgwalters:
e.g. https://github.com/coreos/rpm-ostree/pull/3794

sunfishcode:
Makes sense, thanks!

sunfishcode:
This is now released in 0.25.1.

PR #266, opened 2022-06-28
cap-std: Re-export io-lifetimes

cgwalters:
For similar reasons as we re-export cap-primitives bits.

(Also, the fact that the io-lifetimes semver is different from
 cap-std makes having this even more convenient for users instead
 of juggling two separate dependencies in `Cargo.toml`)


sunfishcode:
Sounds good.

Eventually cap-std will stop using io-lifetimes in its public API, as the traits/types will be available in std, but that change will be a semver bump anyway, so that's no reason not to do this.

sunfishcode:
This is now released in 0.25.2.

PR #268, opened 2022-07-02
Fix statx created times when unavailable

ongardie:
This previously returned `Some(UNIX_EPOCH)` instead of `None` when `statx` did not return a created time.

This is easily tested on Linux against a `tmpfs`. For example, if `/dev/shm` is a `tmpfs`, then `stat /dev/shm` will show "Birth time" as "-" and `TMPDIR=/dev/shm cargo test` will fail the unit tests (without the corresponding fix).

### DIFF
--- a/cap-primitives/src/rustix/fs/metadata_ext.rs
+++ b/cap-primitives/src/rustix/fs/metadata_ext.rs
@@ -3,7 +3,7 @@
 use crate::fs::{FileTypeExt, Metadata, PermissionsExt};
 use crate::time::{Duration, SystemClock, SystemTime};
 #[cfg(any(target_os = "android", target_os = "linux"))]
-use rustix::fs::{makedev, Statx};
+use rustix::fs::{makedev, Statx, StatxFlags};
 use rustix::fs::{RawMode, Stat};
 use std::convert::{TryFrom, TryInto};
 use std::{fs, io};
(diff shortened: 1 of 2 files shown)


tests/fs_additional.rs
```diff
@@ -936,24 +936,44 @@ fn check_metadata(std: &std::fs::Metadata, cap: &cap_std::fs::Metadata) {
...
-    if let Ok(expected) = std.created() {
-        assert_eq!(expected, check!(cap.created()).into_std());
+    match std.created() {
+        Ok(expected) => assert_eq!(expected, check!(cap.created()).into_std()),
+        Err(e) => assert!(
+            cap.created().is_err(),
+            "created time should be error ({}), got {:#?}",
+            e,
+            cap.created()
+        ),
```
ongardie, line 976: This test now requires cap-std to match std's behavior exactly (if and only if). Is that a problem? I could imagine one getting support before the other on new operating systems or architectures, but I don't know if failing tests in those cases is a problem.

cap-primitives/src/rustix/fs/metadata_ext.rs
```diff
@@ -228,7 +228,11 @@ impl MetadataExt {
             permissions: PermissionsExt::from_raw_mode(RawMode::from(statx.stx_mode)),
             modified: system_time_from_rustix(statx.stx_mtime.tv_sec, statx.stx_mtime.tv_nsec as _),
             accessed: system_time_from_rustix(statx.stx_atime.tv_sec, statx.stx_atime.tv_nsec as _),
-            created: system_time_from_rustix(statx.stx_btime.tv_sec, statx.stx_btime.tv_nsec as _),
+            created: if statx.stx_mask & StatxFlags::BTIME.bits() != 0 {
```
ongardie, line 231: The corresponding check in `std` is currently here: https://github.com/rust-lang/rust/blob/master/library/std/src/sys/unix/fs.rs#L461
I don't think it's necessary to check the mask for other fields that have been around for a long time, but I'm not sure.

tests/fs_additional.rs
ongardie: Aha, this failed CI for `musl` with what appears to be a valid timestamp.

cap-primitives/src/rustix/fs/metadata_ext.rs
sunfishcode: It looks like [std does the checks](https://github.com/rust-lang/rust/blob/master/library/std/src/sys/unix/fs.rs#L333) for `STATX_MTIME`, `STATX_ATIME`, and `STATX_CTIME`, so I think cap-std should too.

ongardie: Sorry for the delay on my end. I don't know why that's guarded by `#[cfg(target_pointer_width = "32")]` in std, but it seems harmless to just always check it. I'll update this.

ongardie:
Updated, ready for review again @sunfishcode, thanks! (no hurry)

sunfishcode (approved):
Looks good. Thanks for working on this!

PR #273, opened 2022-08-23
Update to io-lifetimes 1.0.

sunfishcode:
This converts to the new traits, with `From` instead of `IntoFd` and `FromFd`.
    
This currently depends on cap-async-std being disabled, since async-std
doesn't yet have the io-safety trait impls yet.


sunfishcode:
Once https://github.com/async-rs/async-std/pull/1036 lands, we'll be able to re-enable cap-async-std here.

PR #276, opened 2022-09-23
fix: always set `directory` open flag in `open_ambient_dir`

rvolosatovs:
Note, that this is the only way to actually acquire a directory from `AmbientAuthority` when running in Wasmtime.
Fix for that is in-progress at https://github.com/bytecodealliance/wasmtime/pull/4947

All this *should* do is fail if the FD is not a directory. Currently Wasmtime Snapshot 1 implementation incorrectly uses this bit value to determine whether to open the FD in "file mode" or "directory mode". Setting this bit is a workaround for that issue, but it's also just generally a good thing to do.

### DIFF
--- a/cap-primitives/src/rustix/fs/dir_utils.rs
+++ b/cap-primitives/src/rustix/fs/dir_utils.rs
@@ -6,7 +6,7 @@ use std::ops::Deref;
 #[cfg(unix)]
 use std::os::unix::{ffi::OsStrExt, fs::OpenOptionsExt};
 #[cfg(target_os = "wasi")]
-use std::os::wasi::ffi::OsStrExt;
+use std::os::wasi::{ffi::OsStrExt, fs::OpenOptionsExt};
 use std::path::Path;
 #[cfg(racy_asserts)]
 use std::{ffi::OsString, os::unix::ffi::OsStringExt, path::PathBuf};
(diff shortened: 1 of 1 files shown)


rvolosatovs:
CI failures seem unrelated:
```
 [cap-primitives 1.0.0-rc1] For more information about this error, try `rustc --explain E0554`.
error[E0308]: arguments to this function are incorrect
    --> /home/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/rustix-0.36.0-rc1/src/backend/libc/fs/makedev.rs:15:14
     |
15   |     unsafe { c::makedev(maj as i32, min as i32) }
     |              ^^^^^^^^^^ ----------  ---------- expected `u32`, found `i32`
     |                         |
     |                         expected `u32`, found `i32`
     |
note: function defined here
    --> /home/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/libc-0.2.133/src/unix/linux_like/android/mod.rs:2794:20
     |
2794 |     pub {const} fn makedev(ma: ::c_uint, mi: ::c_uint) -> ::dev_t {
     |                    ^^^^^^^
help: you can convert an `i32` to a `u32` and panic if the converted value doesn't fit
     |
15   |     unsafe { c::makedev((maj as i32).try_into().unwrap(), min as i32) }
     |                         +          +++++++++++++++++++++
help: you can convert an `i32` to a `u32` and panic if the converted value doesn't fit
     |
15   |     unsafe { c::makedev(maj as i32, (min as i32).try_into().unwrap()) }
     |                                     +          +++++++++++++++++++++
```

cap-primitives/src/rustix/fs/dir_utils.rs
```diff
@@ -100,15 +100,13 @@ pub(crate) fn canonicalize_options() -> OpenOptions {
...
-    let flags = OFlags::DIRECTORY | target_o_path();
-
     let mut options = fs::OpenOptions::new();
     options.read(true);
 
     #[cfg(not(target_os = "wasi"))]
-    options.custom_flags(flags.bits() as i32);
+    options.custom_flags((OFlags::DIRECTORY | target_o_path()).bits() as i32);
+    #[cfg(target_os = "wasi")]
+    options.directory(true);
```
sunfishcode, line 109: Looks good. Can you preserve the comment about how the reason we need to do this is that we're calling std::fs which doesn't have a `dir_required` function?

sunfishcode:
The `makedev` error in CI is indeed unrelated, and it's now fixed on `main`. Would you mind rebasing this?

rvolosatovs:
@sunfishcode updated, PTAL

cap-primitives/src/rustix/fs/dir_utils.rs
rvolosatovs: done

sunfishcode:
Thanks!

Issue #226, opened 2022-01-24
Use `NtCreateFile` on Windows


sunfishcode:
As @kubkon  observed, `NtOpenFile` provides a more efficient and likely more reliable way to implement `openat`-like behavior on Windows. As an example, Rust is now using it this way in its [`remove_dir_all` implementation](https://github.com/rust-lang/rust/blob/master/library/std/src/sys/windows/fs.rs#L713).

Microsoft officially documents `NtOpenFile` as an [Internal API](https://docs.microsoft.com/en-us/windows/win32/devnotes/calling-internal-apis) which may change between releases or even service packs of Windows, however this particular function, `NtOpenFile`, has apparently been stable for a long time, and since Rust itself is now depending on it, it would seem to be sufficiently stable for cap-std to use as well.

sunfishcode:
Use `NtCreateFile` instead of `NtOpenFile`; see https://github.com/rust-lang/rust/pull/93206.

ChrisDenton:
My current thinking on this has changed since that PR. At the time using any `Nt*` function was controversial so I was eager to be cautious above all else, especially given it's the standard library. Since then I've had more discussion with folks at Microsoft and I now feel much more relaxed about use of `Nt*` functions where it can be justified. So long as they're well documented and have been around for awhile, they are not going to break.

Btw, the [Calling Internal APIs](https://docs.microsoft.com/en-us/windows/win32/devnotes/calling-internal-apis) page, while scary, is somewhat outdated. There now exists an `ntdll.lib` which is distributed with the Windows SDK and allows linking with many ntdll functions without using `GetProcAddress`.  If using a gnu (rather than msvc) toolchain it can not necessarily be assumed that the SDK is installed but if the function is part of the `windows` and `windows-sys` crates then this will be in the import libraries distributed by those crates (and these are also official Microsoft projects).

sunfishcode:
Oh wow, I wasn't previously aware that windows-sys has bindings for [`NtCreateFile`](https://docs.rs/windows-sys/latest/windows_sys/Win32/Storage/FileSystem/fn.NtCreateFile.html) and [`NtOpenFile`](https://docs.rs/windows-sys/latest/windows_sys/Win32/System/WindowsProgramming/fn.NtOpenFile.html).

sunfishcode:
Also, cap-std needs the ability to create files, so it appears we do need `NtCreateFile`.

[...]

PR #277, opened 2022-09-27
On Windows, make `maybe_dir(true)` disable deletion and renaming.

sunfishcode:
Similar to `dir_options()`, on Windows, when `maybe_dir` is set, clear the share flag for deletion and remaining, to protect cap-std's sandboxed path lookup.

And, add a coment to cap-fs-ext's documentation for `maybe_dir`.


sunfishcode:
#226 should obviate this, by avoiding the need to protect cap-std's path lookup from concurrent deletes or renames; this is a temporary workaround.

sunfishcode:
This needs a new rustix release.

PR #282, opened 2022-11-09
Update to io-extras 0.17.

sunfishcode:
This fixes a compilation error on Rust 1.58. Also, add a basic test and README.md declaration for a Rust 1.58 MSRV.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -365,7 +365,7 @@ jobs:
             rust: 1.58
 
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v3
       with:
         submodules: true
     - uses: ./.github/actions/install-rust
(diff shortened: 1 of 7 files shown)


cgwalters (approved):

.github/workflows/main.yml
```diff
@@ -353,6 +353,30 @@ jobs:
...
+  test_msrv:
+    name: Test
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        build: [msrv]
+        include:
+          - build: msrv
+            os: ubuntu-latest
+            rust: 1.58
```
cgwalters, line 365: In some other actions we scrape this value out of `Cargo.toml`...see e.g. https://github.com/coreos/bootupd/blob/e167ea9e7c2539654aba94f903061718ac499bc1/.github/workflows/rust.yml#L57

If you choose to keep this as is, probably worth adding a note in `Cargo.toml` to bump the `rust-version` there too.

Thanks for digging into this!

cgwalters (approved):

.github/workflows/main.yml
sunfishcode: Ah, I hadn't added a rust-version field yet. I've now added it, with a comment!

PR #287, opened 2023-01-03
tempfile: Add basic test for `new_anonymous`

cgwalters:
On general principle, but also specifically because I think this may not work on Windows; came out of some discussion in https://github.com/containers/image/pull/1769

### DIFF
--- a/cap-tempfile/src/tempfile.rs
+++ b/cap-tempfile/src/tempfile.rs
@@ -230,6 +230,16 @@ mod test {
         panic!("Could not determine process umask")
     }
 
+    /// Older Windows versions don't support removing open files
+    fn os_supports_unlinked_tmp(d: &Dir) -> bool {
+        if cfg!(not(windows)) {
+            return true;
+        }
+        let name = "testfile";
+        let _f = d.create(name).unwrap();
+        d.remove_file(name).and_then(|_| d.create(name)).is_ok()
+    }
+
     #[test]
     fn test_tempfile() -> io::Result<()> {
         use crate::ambient_authority;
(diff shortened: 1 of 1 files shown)


cgwalters:
```
 ---- tempfile::test::test_tempfile stdout ----
Error: Os { code: 303, kind: Uncategorized, message: "The file cannot be opened because it is in the process of being deleted." }
```

OK, yep; I messed up here in adding this, sorry!  It was a last minute addition, IIRC thinking of it as a way to eventually expose the Linux `O_TMPFILE | O_EXCL` semantics.  Not adding a basic unit test was the second mistake.

Anyways, we have an API that only works on Unix.  Should we:

- Add a note in the documentation that this API will always fail on Windows currently
- Make it `#[cfg(unix)]` but without breaking semver (assuming no one is using it today)
- Queue it up for a future semver break?  (probably never)
- Something else?

ChrisDenton:
You may want test this on windows-latest (i.e. temporarily disable the windows-2019 test).

ChrisDenton:
I'm uncertain of how this test ultimately boils down to syscalls but the behaviour of `DeleteFileW` has changed. It now uses "POSIX semantics" by default. I.e. it removes the file from the directory tree immediately instead of the win32 semantics of waiting until the last file handle is closed.

cgwalters:
> I'm uncertain of how this test ultimately boils down to syscalls but the behaviour of DeleteFileW has changed. It now uses "POSIX semantics" by default. 

Ah, neat.  I'm sure that change was motivated by lots of similar code issues.

OK, and now that I look closer I see CI is already running `windows-latest`, we just got hit by the fail-fast semantics.  I added a temporary commit which drops windows-2019.

If CI passes on windows-latest, then ISTM the clearly right thing to do is to document the API only works on newer Windows and move on.

cgwalters:
@ChrisDenton Do you know what's the most convenient way to dynamically detect the appropriate Windows version and skip the `new_anonymous()` test if it's too old?

ChrisDenton:
That's actually a bit complicated. The easiest thing to do would be to feature detect by doing. E.g. creating a file (keeping it open), deleting it and then creating a new file at the same path. Otherwise you'd need to figure out [RtlVerifyVersionInfo](https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/wdm/nf-wdm-rtlverifyversioninfo) (higher level functions exist but they lie unless a compatibility shim is disabled),

cgwalters:
OK thanks, I took a crack at that.

PR #288, opened 2023-01-03
tempfile: Use `O_TMPFILE | O_EXCL` for `new_anonymous()`

cgwalters:
The addition of `new_anonymous()` was a bit of a last minute thing when I was working on the code originally.  I'd meant to have it expose the Linux semantics for `O_TMPFILE | O_EXCL`, which means the file can never be linked into place. In theory (but I don't think in practice) this could allow e.g. the underlying filesystem and storage stack to optimize things knowing that the data will not persist across a reboot for example.


sunfishcode:
Looks good!

PR #299, opened 2023-03-15
Add a `Pool::insert_ip_net_any_port` function.

sunfishcode:
Add a function to `Pool` to allow inserting a network that can accept any port.


sunfishcode:
This also now adds support for port ranges. And `std::net` doesn't do IANA port name lookups, so this removes the TODO for that as well.

PR #312, opened 2023-04-10
Do a `uname` version check before using `statx` on Android.

sunfishcode:
Similar to #197, do a `uname` version check before using `statx` on Android, since Android's seccomp configuration kills the process if any unrecognized system calls are used.

Fixes #311.

PR #316, opened 2023-04-20
Avoid calling `openat2` on Android in more places.

sunfishcode:
Disable the `openat2` code entirely on Android with a `cfg`, and adjust other code to avoid calling it.

This fixes some `openat2` calls that remained after #312.

### DIFF
--- a/cap-primitives/src/rustix/linux/fs/mod.rs
+++ b/cap-primitives/src/rustix/linux/fs/mod.rs
@@ -1,23 +1,36 @@
+#[cfg(target_os = "linux")]
 mod canonicalize_impl;
+#[cfg(target_os = "linux")]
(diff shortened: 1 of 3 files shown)


eddyb (approved):
Tested by:
1. removing my local workaround
   - outcome: app crashes with the syscall SECCOMP violation (as expected)
2. switching to this PR using:
   ```toml
   [patch.crates-io]
   cap-primitives = { git = "https://github.com/bytecodealliance/cap-std", branch = "sunfishcode/android-openat2" }
   ```
   - outcome: app works again!

sunfishcode:
Thanks!

sunfishcode:
This is now released in cap-std et al 1.0.14.

PR #327, opened 2023-08-24
Add timezone.

cdmurph32:
This code only fetches the name of the timezone. The determination of utc offset is left to the runtime. 

Issue:
https://github.com/bytecodealliance/wasmtime/issues/4274

WASI Clocks proposal:
https://github.com/WebAssembly/wasi-clocks/pull/48

Relate wasmtime PR:
https://github.com/bytecodealliance/wasmtime/pull/6899

### DIFF
--- a/cap-primitives/Cargo.toml
+++ b/cap-primitives/Cargo.toml
@@ -20,7 +20,6 @@ maybe-owned = "0.3.4"
 fs-set-times = "0.20.0"
 io-extras = "0.18.0"
 io-lifetimes = { version = "2.0.0", default-features = false }
-iana-time-zone = "0.1.57"
 
 [dev-dependencies]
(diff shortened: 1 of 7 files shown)


sunfishcode:
I apologize for the slow response here!

The cap-std crate is intended to focus on features which are provided by Rust's `std`. What would you think about moving this code to the cap-time-ext crate, which also lives in this repository? We could also move the implementation from cap-primitives into cap-time-ext, so that users that don't need it wouldn't need the `iana-time-zone` dependency.

cdmurph32:
No apology necessary. I just figured out how to get it working end-to-end with wasmtime this morning. I didn't even know about the component adapter.

sunfishcode:
Thanks!

PR #329, opened 2023-09-13
Fix stats where `st_dev` is negative

alexcrichton:
This is intended to be a fix for bytecodealliance/wasmtime#6978 where some platforms use a signed integer for the `dev_t` type and it looks like some `st_dev` fields can indeed be negative.


alexcrichton:
I've pushed up a few more commits here to fix some issues on CI as well and I think this should be good to go now

sunfishcode:
Looks good, thanks!

PR #333, opened 2023-09-18
Fix DirEntryInner::is_same_file at a mountpoint

### LINKED ISSUE #330: file_path_by_searching does not work across mountpoint boundaries
file_path_by_searching tries to identify a directory's pathname by iterating through its parent's children and matching the dev and ino fields.  The problem is that if the directory in question is a file system root, then its ino won't match.  Readdir will show the ino of the underlying mountpoint, whereas the directory's own ino method will show the ino of its root directory.

Steps to Reproduce
================
On a non-Linux, non-Darwin system (tested on FreeBSD 15.0-CURRENT), set TMPDIR to a file system's mountpoint.  This is the default on a ZFS-root system, where /tmp is a separate ZFS file system.
```
cd cap-primitives
cargo test --lib fs::dir_paths
```

asomers:
If a direntry comes from a mountpoint's parent, then its ino will be the ino of the underlying directory, not the mounted file system's root directory.  So ignore the direntry's ino, and just look at the ino in its metadata.

Fixes #330

### DIFF
--- a/cap-primitives/src/rustix/fs/dir_entry_inner.rs
+++ b/cap-primitives/src/rustix/fs/dir_entry_inner.rs
@@ -69,7 +69,8 @@ impl DirEntryInner {
 
     #[inline]
     pub(crate) fn is_same_file(&self, metadata: &Metadata) -> io::Result<bool> {
-        Ok(self.ino() == metadata.ino() && self.metadata()?.dev() == metadata.dev())
+        let self_md = self.metadata()?;
+        Ok(self_md.ino() == metadata.ino() && self_md.dev() == metadata.dev())
     }
 
     fn file_name_bytes(&self) -> &OsStr {


asomers:
The test failure is on OSX, and I don't have a Mac to debug it on.  Could one of the maintainers please help?

sunfishcode:
I apologize for the slow response here, I lost track of this. The CI was restarted so it now passed.

I think if `DirEntry::is_same_file` were a public API, I might have different thoughts about this, because it might sometimes be desirable to compare with the ino within the underlying directory rather than any filesystem that might happen to be mounted on top of it. However, this isn't a public API, and it's currently only used for these `file_path_by_searching` and `remove_open_dir_by_searching` functions which don't need that behavior, so it seems good.

PR #334, opened 2023-09-18
Fix the tty_path test on FreeBSD

asomers:
On FreeBSD, ttyname(3) may not return exactly the same name as was used to open the file descriptor.  For example, "/dev/tty" is an alias for whatever the real tty device is.  But paths like "/dev/pts/0" aren't aliases.  If the user can open them, ttyname() will always return their own name.

Note that the test will still fail on non-Linux, non-Darwin due to an unrelated Rustix bug.  See
https://github.com/bytecodealliance/rustix/pull/832


sunfishcode:
Thanks!

PR #343, opened 2024-01-03
Use custom traits instead of libstd's `MetadataExt` and `PermissionsExt`.

### LINKED ISSUE #342: Stop implementing std's `MetadataExt`
Rust is looking to seal `std::os::unix::fs::MetadataExt`:

https://github.com/rust-lang/rust/pull/118916

As in https://github.com/bytecodealliance/cap-std/pull/271 (which was before cap-std 1.0), cap-std can switch to defining its own copies of these traits. And since cap-std seems to be the only crate in the ecosystem that is affected, it should probably do this, to unblock std evolution.

However, this is a semver-incompatible change, so it'll require bumping cap-std to 3.0.

@cgwalters, who I know is particularly impacted by semver bumps

sunfishcode:
Define our own copy of `MetadataExt` and `PermissionsExt` and change the code to use them instead of `std::os::*::fs::MetadataExt` and `std::os::*::fs::PermissionsExt'.

In general, we should be moving away from using std's `Ext` traits like this, as they weren't meant to be used in this way.

Fixes #342.

### DIFF
--- a/cap-primitives/src/fs/metadata.rs
+++ b/cap-primitives/src/fs/metadata.rs
@@ -1,4 +1,4 @@
-use crate::fs::{FileType, ImplFileTypeExt, MetadataExt, Permissions};
+use crate::fs::{FileType, ImplFileTypeExt, ImplMetadataExt, Permissions};
 use crate::time::SystemTime;
 use std::{fs, io};
 
@@ -18,7 +18,7 @@ pub struct Metadata {
     pub(crate) modified: Option<SystemTime>,
     pub(crate) accessed: Option<SystemTime>,
     pub(crate) created: Option<SystemTime>,
-    pub(crate) ext: MetadataExt,
+    pub(crate) ext: ImplMetadataExt,
 }
 
 #[allow(clippy::len_without_is_empty)]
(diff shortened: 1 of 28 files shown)


cap-primitives/src/fs/metadata.rs
```diff
@@ -198,129 +198,191 @@ impl Metadata {
...
+    fn mtime_nsec(&self) -> i64;
+    /// Returns the last status change time of the file, in seconds since Unix Epoch.
+    fn ctime(&self) -> i64;
+    /// Returns the last status change time of the file, in nanoseconds since [`ctime`].
+    fn ctime_nsec(&self) -> i64;
+    /// Returns the block size for filesystem I/O.
+    fn blksize(&self) -> u64;
+    /// Returns the number of blocks allocated to the file, in 512-byte units.
+    fn blocks(&self) -> u64;
+    #[cfg(target_os = "vxworks")]
```
cgwalters, line 238: It seems easy for us to somehow miss future extensions to these traits in std, especially slightly more obscure OSes.  This seems to be the main downside of this.  But eh, in the end if there's some obscure new API and we don't notice, it can just be added.

cgwalters (approved):
Seems sane to me

cap-primitives/src/fs/metadata.rs
sunfishcode: Yeah. All of cap-std already faces this challenge; it's just the nature of what cap-std being what it is.

PR #354, opened 2024-05-14
Allow read_link to read absolute paths

### LINKED ISSUE #353: read_link should be allow absolute paths
The `read_link` function will return an error if the link's destination is an absolute path.  There are three problems with this:
* It _won't_ return an error if the link's destination is a relative path that uses `../` to point outside of the sandbox.  That's inconsistent.
* Sometimes a symlink with an absolute path _doesn't_ escape the sandbox.  For example `/sandbox/link -> /sandbox/target`.
* More importantly, sometimes it's important to read a link that points outside of the sandbox.  For example, a backup application could copy all of the files from the root directory of one computer to a subdirectory of another.  Symlinks would be broken, but would work again after a restore operation.  The application that performs the restore would need to be able to read link targets, even if they point outside of the root.  Or, a jail file system could contain absolute symlinks that resolve correctly for a jailed process, but not for an unjailed one.  But it would still sometimes be useful for an unjailed process to read the links.

I suggest simply removing the absolute path check.

asomers:
Previously it would return an error if the link's destination were absolute.  But there were three problems with this:

* It wouldn't return an error if the link's destination were a relative path that uses ../ to point outside of the sandbox. That's inconsistent.
* Sometimes a symlink with an absolute path doesn't escape the sandbox. For example /sandbox/link -> /sandbox/target.  But read_link wouldn't allow that.
* More importantly, sometimes it's important to read a link that points outside of the sandbox. For example, a backup application could copy all of the files from the root directory of one computer to a subdirectory of another. Symlinks would be broken, but would work again after a restore operation. The application that performs the restore would need to be able to read link targets, even if they point outside of the root. Or, a jail file system could contain absolute symlinks that resolve correctly for a jailed process, but not for an unjailed one. But it would still sometimes be useful for an unjailed process to read the links.

Fixes #353

### DIFF
--- a/cap-primitives/src/fs/read_link.rs
+++ b/cap-primitives/src/fs/read_link.rs
@@ -1,7 +1,7 @@
 //! This defines `read_link`, the primary entrypoint to sandboxed symbolic link
 //! dereferencing.
 
-use crate::fs::{errors, read_link_impl};
+use crate::fs::read_link_impl;
 #[cfg(racy_asserts)]
 use crate::fs::{map_result, read_link_unchecked, stat, FollowSymlinks};
 use std::path::{Path, PathBuf};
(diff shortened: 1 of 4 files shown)


asomers:
Superseded by #356

PR #356, opened 2024-05-15
Add Dir::{read_link_contents,symlink_contents} methods

asomers:
These allow creating and readling symlinks whose target may be an absolute path.  There is no security risk because security is as always provided by the OS; the OS's openat implementation will refuse to follow a symlink whose target is an absolute path.

Fixes #354


sunfishcode:
Looks good, thanks!

sunfishcode:
This is now published in cap-std 3.1.0.

PR #357, opened 2024-06-28
fs_utf8: Add `as_cap_std()`

cgwalters:
I'm looking at adding support in https://github.com/coreos/cap-std-ext/ for `fs_utf8`, and doing so really wants access to the currently internal `cap_std` member in order to share code - in the same way basically all the implementation functions here just delgate to the underlying `Dir`.

I changed one usage to call this helper.


sunfishcode:
Makes sense, thanks!

sunfishcode:
This is now released in cap-std 3.2.0.

PR #360, opened 2024-07-09
tempfile: Delegate `tempdir::new` to base version

cgwalters:
I was looking at the code here for an unrelated reason, and realized that there's no reason for us to duplicate the code.

Because a Dir instance doesn't have a name accessible to the calling code, we just delegate to the non-UTF8 base version. In some presumably very unusual situations the tempdir may have a non-UTF8 name, but that's fine.

### DIFF
--- a/cap-tempfile/src/utf8.rs
+++ b/cap-tempfile/src/utf8.rs
@@ -2,12 +2,13 @@
 //!
 //! TODO: This whole scheme is still under development.
 
+#[cfg(test)]
 use camino::Utf8PathBuf;
 use cap_std::fs_utf8::Dir;
+#[cfg(test)]
+use std::env;
 use std::ops::Deref;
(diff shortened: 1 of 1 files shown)


cap-tempfile/src/utf8.rs
```diff
@@ -39,27 +47,11 @@ impl TempDir {
     /// This function makes use of ambient authority to access temporary
     /// directories.
     pub fn new(ambient_authority: AmbientAuthority) -> io::Result<Self> {
-        let system_tmp: Utf8PathBuf = env::temp_dir()
-            .try_into()
-            .expect("temporary directory path should be valid UTF-8");
```
cgwalters, line 44: Notably this `panic` goes away; it was never necessary.

sunfishcode:
Nice!

sunfishcode:

PR #361, opened 2024-07-11
[draft] Add borrowing for fs_utf8::Dir

cgwalters:
I'd like to start using `fs_utf8::Dir`, but not in a "big bang" fashion where I wholesale convert the entire codebase. It feels normal and natural to have a non-utf8 `Dir` instance and maybe sometimes just want to "view" it as utf8.

At an efficiency level, having to create a new owned fd for this feels unnecessary.

This sketches out an implementation where we can also support borrowing an existing fd.

This doesn't compile, but before I do a lot more typing on this - consider it an RFC. Is this a good direction?

### DIFF
--- a/cap-std/src/fs_utf8/dir.rs
+++ b/cap-std/src/fs_utf8/dir.rs
@@ -11,6 +11,7 @@ use io_lifetimes::AsFilelike;
 use io_lifetimes::{AsFd, BorrowedFd, OwnedFd};
 #[cfg(windows)]
 use io_lifetimes::{AsHandle, BorrowedHandle, OwnedHandle};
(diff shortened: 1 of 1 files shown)


cgwalters:
Oh yeah, I guess things like

```
#[cfg(not(windows))]
impl From<Dir> for OwnedFd {
    #[inline]
    fn from(dir: Dir) -> OwnedFd {
        dir.cap_std.into()
    }
}
```

can't work anymore...ugh, would need a new `BorrowedDir` or something.

cgwalters:
OK, trying https://github.com/bytecodealliance/cap-std/pull/362 instead

PR #362, opened 2024-07-11
dir: Add convenience functions for accessing UTF-8 variants

cgwalters:
I'd like to make use of the fs_utf8 bits in some of my code, but doing so is tricky as switching `Dir` types quickly becomes "infectious" across the codebase and forces a larger conversion all at once.

Adding these these two convenience APIs on `Dir` (the non-UTF8 version) I think greatly improve the ergonomics for the common cases where I may want to *view* an existing `Dir` entries (without creating a new file descriptor) and iterate over its entries, and to conveniently open a child dir as a utf-8 version.

There are more methods we could add, but these feel like a useful start to me.

### DIFF
--- a/cap-std/src/fs/dir.rs
+++ b/cap-std/src/fs/dir.rs
@@ -1,6 +1,8 @@
 #[cfg(target_os = "wasi")]
 use crate::fs::OpenOptionsExt;
 use crate::fs::{DirBuilder, File, Metadata, OpenOptions, ReadDir};
+#[cfg(feature = "fs_utf8")]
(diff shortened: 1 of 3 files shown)


cgwalters:
ref e.g. what camino does:

- https://docs.rs/camino/latest/camino/struct.Utf8Path.html#method.read_dir
- https://docs.rs/camino/latest/camino/struct.Utf8Path.html#method.read_dir_utf8

sunfishcode:
Makes sense, thanks!

sunfishcode:

PR #363, opened 2024-07-17
tempdir: Add a `persist` method

cgwalters:
I have a use case where I want the equivalent of `into_path()`.

### DIFF
--- a/cap-tempfile/src/lib.rs
+++ b/cap-tempfile/src/lib.rs
@@ -101,6 +101,16 @@ impl TempDir {
         Err(Self::already_exists())
     }
 
+    /// Make this directory persistent.
+    ///
+    /// This corresponds to [`tempfile::TempDir::into_path`], but returns a [`Dir`].
+    ///
+    /// [`tempfile::TempDir::into_path`]: https://docs.rs/tempfile/latest/tempfile/struct.TempDir.html#method.into_path
+    /// [`Dir`]: https://docs.rs/cap-std/latest/cap_std/fs/struct.Dir.html
+    pub fn into_dir(mut self) -> io::Result<Dir> {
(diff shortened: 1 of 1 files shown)


cap-tempfile/src/lib.rs
```diff
@@ -101,6 +101,15 @@ impl TempDir {
...
     }
 
+    /// Make this directory persistent.
+    ///
+    /// This corresponds to [`tempfile::TempDir::into_path`].
+    ///
+    /// [`tempfile::TempDir::into_path`]: https://docs.rs/tempfile/latest/tempfile/struct.TempDir.html#method.into_path
+    pub fn persist(mut self) -> io::Result<Dir> {
+        Ok(self.dir.take().unwrap())
+    }
```
sunfishcode, line 111: Would it make sense to call this function `into_dir`, for closer analogy with the tempfile API?

cap-tempfile/src/lib.rs
```diff
@@ -101,6 +101,15 @@ impl TempDir {
         Err(Self::already_exists())
     }
 
+    /// Make this directory persistent.
+    ///
+    /// This corresponds to [`tempfile::TempDir::into_path`].
```
sunfishcode, line 106: Could you add ", except that it returns a `Dir` instead of a `PathBuf`" to this comment?

sunfishcode (commented):
Overall looks good, just a few minor comments.

cap-tempfile/src/lib.rs
cgwalters: Sure, done!

sunfishcode:
Thanks!

sunfishcode:

PR #364, opened 2024-07-24
fs_utf8/Dir: impl AsRef<crate::fs::Dir>

cgwalters:
This allows an API like:
```
fn foo(d: impl AsRef<Dir>) {
    ...
}
```

to accept either a `Dir` or a `DirUtf8`. I don't know that I'll do this much, but I think it can make sense.


sunfishcode:
Seems reasonable.

sunfishcode:

PR #369, opened 2024-10-29
[cap-primitives] add support for illumos

sunshowers:
Get the cap-primitives library building and passing all tests on illumos. There were just a couple of minor tweaks required.

I also ran into an issue where the minimum version of `ipnet` required was too old, so I took the liberty to bump it to 2.5.0 which has `IpNet::new`.

The note in `dir_entry.rs` is copied over from Rust's std: https://doc.rust-lang.org/std/fs/struct.DirEntry.html#method.file_type

### DIFF
--- a/cap-primitives/src/rustix/fs/mod.rs
+++ b/cap-primitives/src/rustix/fs/mod.rs
@@ -142,10 +142,13 @@ fn tty_path() {
     #[cfg(unix)]
     use std::os::unix::fs::FileTypeExt;
 
-    // On FreeBSD, /dev/{tty,stdin,stdout,stderr} are aliases to different real
-    // devices.
     let paths: &[&str] = if cfg!(target_os = "freebsd") {
+        // On FreeBSD, /dev/{tty,stdin,stdout,stderr} are aliases to different
+        // real devices.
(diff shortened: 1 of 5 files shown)


cap-primitives/src/rustix/fs/mod.rs
```diff
@@ -142,10 +142,13 @@ fn tty_path() {
     #[cfg(unix)]
     use std::os::unix::fs::FileTypeExt;
 
-    // On FreeBSD, /dev/{tty,stdin,stdout,stderr} are aliases to different real
-    // devices.
     let paths: &[&str] = if cfg!(target_os = "freebsd") {
+        // On FreeBSD, /dev/{tty,stdin,stdout,stderr} are aliases to different
+        // real devices.
         &["/dev/ttyv0", "/dev/pts/0"]
+    } else if cfg!(target_os = "illumos") {
```
sunshowers, line 149: I wasn't sure about Solaris here so I left this as illumos. The other bits I'm reasonably sure are okay on Solaris.

sunfishcode:
Looks good, thanks!

sunfishcode:
This is now released in cap-std 3.4.0.

sunshowers:
Thank you!

PR #371, opened 2024-11-01
Add missing reserved file names

nathaniel-daniel:
The Windows open primitive currently rejects some reserved file names like CON and LPT1 if the user attempts to open them. However, this library does not include reserved file names with superscript digits in this list. This PR adds these missing reserved file names.


sunfishcode:
Thanks!

PR #388, opened 2025-04-16
Fix a panic when opening a directory with `truncate(true)`

alexcrichton:
This is intended to address bytecodealliance/wasmtime#10577


sunfishcode:
Thanks!

PR #407, opened 2025-12-29
cap-directories: Support XDG_STATE_HOME

mattias-p:
Hi guys! `cap-std` looks neat! I'm looking into using it, but I need to put stuff into [XDG_STATE_HOME]. Here's an update to support it.

To achieve this I replaced the dependency on [directories-next] with [directories]. This change seems reasonable to me since the latter is maintained again, while the former is not.

[XDG_STATE_HOME]: https://specifications.freedesktop.org/basedir/latest/#basics
[directories-next]: https://crates.io/crates/directories-next
[directories]: https://crates.io/crates/directories


sunfishcode:
The PR looks good. Thanks!

PR #413, opened 2026-04-16
Move FreeBSD CI testing from Cirrus to Github Workflows

asomers:
Because Cirrus Labs will soon shut down.


asomers:
The test failure was introduced by 4e8ba003ad00920f21edcb2d126cc15ed302508d .  And it's not a serious failure; just a mismatched expectation for an error message.  Personally, I think the `Capabilities insufficient (os error 93)` error is adequate.
It seems that cap-std hasn't actually been running CI on FreeBSD lately.  The .cirrus.yml file existed, but I don't see any recent checks.  Did the Cirrus app get disabled at some point?

sunfishcode:
Thanks!

PR #414, opened 2026-05-11
Bring back `cap-async-std` as `cap-tokio`

### LINKED ISSUE #408: Replace cap-async-std
cap-async-std has been removed without a replacement (see https://github.com/bytecodealliance/cap-std/pull/399), because async-std is no longer maintained.
Now there is no way of using async code with this crate.

It would make sense to add either a:
- cap-tokio (I would prefer this one)

or a
- cap-smol

crate to this repository to retain the async functionality of this project.

Ericson2314:
Fixes #408

I don't know enough about to `tokio` to really be the person that does this, but it does compile for me.

This was based on reverting commit ecc886dfbf9c2f7438ef06993abc12c8ca19f320.

### DIFF
--- a/cap-fs-ext/Cargo.toml
+++ b/cap-fs-ext/Cargo.toml
@@ -14,9 +14,12 @@ edition = "2021"
 
 [dependencies]
 arf-strings = { version = "0.7.0", optional = true }
+cap-tokio = { path = "../cap-tokio", optional = true, version = "4.0.2" }
 cap-std = { path = "../cap-std", optional = true, version = "4.0.2" }
 cap-primitives = { path = "../cap-primitives", version = "4.0.2" }
 io-lifetimes = { version = "3.0.1", default-features = false }
+async-std = { version = "1.13.0", features = ["io_safety", "attributes"], optional = true }
+async-trait = { version = "0.1.42", optional = true }
 camino = { version = "1.0.5", optional = true }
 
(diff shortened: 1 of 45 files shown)


cap-fs-ext/Cargo.toml
```diff
@@ -14,16 +14,22 @@ edition = "2021"
...
+async-std = { version = "1.13.0", features = ["io_safety", "attributes"], optional = true }
+async-trait = { version = "0.1.42", optional = true }
 camino = { version = "1.0.5", optional = true }
 
 [features]
 default = ["std"]
 fs_utf8 = ["cap-std/fs_utf8", "camino"]
 arf_strings = ["cap-std/arf_strings", "fs_utf8", "arf-strings"]
 std = ["cap-std"]
+async_std = ["cap-tokio", "async-std", "io-lifetimes/async-std", "async-trait"]
```
0xIO32, line 30: I think it makes more sense, if this feature is called tokio now. Also, async-std is discontinued, which is the reason for this effort. There is no reason why it should be a dependency here.

0xIO32 (changes requested):

cap-fs-ext/Cargo.toml
Ericson2314: Yup that's right. This is an oversight.

PR #415, opened 2026-07-02
fs: rename_exclusive (renameat2 + Windows impl)

### LINKED ISSUE #406: Support for renameat2?
On Linux, `cap-std` implements `Dir::rename` via `rustix::fs::renameat`, which is the `renameat` function.

Since Linux 3.15, there is another version of this call, `renameat2`, which supports an additional `flags` argument. `rustix` already supports this function as `rustix::fs::renameat_with`.

Could `cap-std` add support for this function as well?

## Additional Context

The external `cap-std-ext` crate implements atomic writes by writing a temporary file and renaming it in place. However, it is not possible to atomically create a file in this way. 

`renameat2` supports the `RENAME_NOREPLACE` flag, which ensures that no existing file is overwritten by the rename. As such, an `atomic_create` could be implemented based on this change.

joshuamegnauth54:
Closes: #406

POSIX's `rename` atomically replaces targets. However, if a caller wants to optionally replace a target, they have to check if the target exists and then rename onto the target. This sequence of events is subject to TOCTOU where the target may be created between the check and the call to rename().

`renameat2` for some of the Unixes and MoveFileExW on Windows both support atomically checking if a target exists on rename. `renameat2` is supported on Linux, FreeBSD 16 (not exposed yet in libc), Redox, and macOS.

---

I tested this on Linux **but not** Windows. I wrote the Windows code by looking up examples and using Microsoft's documentation. `rename_excl_unchecked` reuses the existing `rename_unchecked` code but calls `renameat2` instead on Unix. I copied Rustix's feature gates for Unix. FreeBSD 16-CURRENT supports `renameat2`, so it can be enabled whenever it's released and Rust's libc _and_ Rustix gains wrappers for it.

The Windows version supports moving files across volumes. It seemed like the correct thing to do and `MoveFileExW` supports it natively. The problem is that file security isn't moved/copied, so it has to manually stored and then restored. The Windows code is significantly shorter without this feature. It's only one call to `MoveFileExW` without it. :joy_cat: 

Anyway, I'm relying on CI to yell at me if the Windows code is incorrect. I tried to be really careful as well as write a lot of comments to document my thinking. The Unix code is straightforward, though.

Windows sources:

* [MoveFileExW](https://learn.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-movefileexw)
* [File Security and Access Rights](https://learn.microsoft.com/en-us/windows/win32/fileio/file-security-and-access-rights)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -43,6 +43,7 @@ version = ">=0.60, <0.62"
 features = [
     "Win32_Storage_FileSystem",
     "Win32_Foundation",
+    "Win32_Security_Authorization",
     "Win32_System_Ioctl",
     "Win32_System_SystemServices",
 ]
--- a/cap-fs-ext/src/dir_entry_ext.rs
+++ b/cap-fs-ext/src/dir_entry_ext.rs
@@ -1,6 +1,6 @@
-use cap_primitives::fs::Metadata;
 #[cfg(windows)]
 use cap_primitives::fs::_WindowsDirEntryExt;
+use cap_primitives::fs::Metadata;
 use std::io;
 
 /// Extension trait for `DirEntry`.
--- a/cap-primitives/src/fs/mod.rs
+++ b/cap-primitives/src/fs/mod.rs
@@ -64,16 +64,16 @@ pub use canonicalize::canonicalize;
 pub use copy::copy;
 pub use create_dir::create_dir;
 pub use dir_builder::*;
-pub use dir_entry::DirEntry;
 #[cfg(windows)]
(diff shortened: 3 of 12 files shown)


joshuamegnauth54:
I currently don't use Windows, so I don't know its semantics. Apparently, volume to volume (e.g. `C:\` to `D:\`) renames are generally not supported. Python's `os` module simply calls `MoveFileExW`. Rust's standard library also doesn't support volume to volume moves. I can vastly simplify the Windows implementation to follow Rust and Python if you'd prefer.

PR #418, opened 2026-08-17
Run `cargo fmt`.

sunfishcode:


cgwalters (approved):
Obviously sane, though I think best practice is to gate CI on `cargo fmt` to avoid drift.

sunfishcode:
cap-std's CI does already have a rustfmt gate; I suspect rustfmt changed slightly.

